PR #13, opened 2019-05-02
solution

cheerfulGloom:


src/index.html
```diff
@@ -13,6 +13,299 @@
...
+      </header>
+      <div class="catalog">
+      <div class="card" data-qa="card">
+        <div>
+          <img class="imac" src="./images/imac.jpeg" alt="iMac">
+        </div>
+        <div class="title">
+          <h2>Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
+        </div>
+        <div>
```
gestap: Maybe better use only one block?
like this:
``` html
<div class="title">
        <h2>Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)</h2>
        <p class="productCode">Код товара: 195434</p>
</div>
```

src/index.html
```diff
@@ -13,6 +13,299 @@
...
+              <img class="star" src="images/star-active.svg" alt="star-active">
+              <img class="star" src="images/star-active.svg" alt="star-active">
+              <img class="star" src="images/star-active.svg" alt="star-active">
+              <img class="star" src="images/star.svg" alt="star">
+          </div>
+          <div>
+            <p class="feedback">Отзывов: 5</p>
+          </div>
+        </div>
+        <div class="priceWord">
```
gestap: And here too

src/style.css
```diff
@@ -0,0 +1,177 @@
...
+
+header {
+  display: flex;
+  height: 60px;
+  margin: 0 auto;
+  padding: 0 50px;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  font-family: Roboto;
+  font-style: normal;
```
gestap: You write `font-family` and `font-style` in `body`. No need to duplicate it.

src/style.css
```diff
@@ -0,0 +1,177 @@
...
+
+.is-active {
+  display: flex;
+  flex-direction: column;
+}
+
+.is-active::after {
+  content: "";
+  margin-top: 20px;
+  padding: 2px;
```
gestap: In this case use `height`.

src/index.html
```diff
@@ -13,6 +13,267 @@
   <link rel="stylesheet" href="./style.css">
```
gestap: You don't connected link for a font.

gestap: after you push new code, run deploy.

PR #18, opened 2019-05-03
Fs apr19 vadims

Kresh24107:


src/style.css
```diff
@@ -0,0 +1,214 @@
...
+  left: 0;
+  bottom: 0;
+  background: #00aedd;
+  border-radius: 8px;
+}
+
+/* styles for product card */
+
+.link {
+  text-decoration: none;
```
gestap: You duplicated this style.
Maybe better add this to `a`. It will delete underscore from all `a`.

src/style.css
```diff
@@ -0,0 +1,214 @@
...
+  align-items: center;
+}
+
+.nav-link {
+  text-transform: uppercase;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  position: relative;
+  text-decoration: none;
```
gestap: !

src/style.css
```diff
@@ -0,0 +1,214 @@
...
+  border-radius: 5px;
+  margin-bottom: 51px; /* magic code here! */
+}
+
+.wrap-img {
+  padding: 31px 20px;
+  width: 158px;
+}
+
+.wrap-img-2 { /* magic code here! */
```
gestap: Do we need this magic?) 
In browser I see nothing

src/style.css
```diff
@@ -0,0 +1,214 @@
...
+  padding: 30px 20px 35px; /* magic code here! */
+  height: 132px;
+}
+
+.product-2 { /* magic code here! */
```
gestap: Also this

PR #20, opened 2019-05-03
Solution

AlexHladun:


readme.md
```diff
@@ -45,5 +45,5 @@ Screenshot of the catalog page:
 
 **Result**
 
-- [UPDATE DEMO LINK](https://<your_account>.github.io/<repo_name>/)
-- [UPDATE TEST REPORT LINK](https://<your_account>.github.io/<repo_name>/report/html_report/)
+- [UPDATE DEMO LINK](https://AlexHladun.github.io/layout_catalog
+- [UPDATE TEST REPORT LINK](https://AlexHladun.github.io/layout_catalog/report/html_report/)
```
gestap: link didn't work

src/style.css
```diff
@@ -0,0 +1,164 @@
...
+}
+
+header img {
+  width: 40px;
+  height: 40px;
+}
+
+nav {
+  left: 397px;
+  right: 50px;
```
gestap: `right` and `left` works only with `position`

src/style.css
```diff
@@ -0,0 +1,164 @@
...
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 14px;
+  text-transform: uppercase;
+  display: flex;
+  align-items: center;
+}
+
+a {
```
gestap: better add `a` class

readme.md
```diff
@@ -45,5 +45,5 @@ Screenshot of the catalog page:
 
 **Result**
 
-- [UPDATE DEMO LINK](https://<your_account>.github.io/<repo_name>/)
-- [UPDATE TEST REPORT LINK](https://<your_account>.github.io/<repo_name>/report/html_report/)
+- [UPDATE DEMO LINK](https://AlexHladun.github.io/layout_catalog/)
```
gestap: gh-page and code below are different.
maybe you forgot run `deploy`.

src/style.css
```diff
@@ -0,0 +1,164 @@
+body {
```
gestap: Знімок екрана  о 10 29 37 дп

gestap: all content on the left side

src/style.css
gestap: !

src/style.css
gestap: !

src/style.css
```diff
@@ -0,0 +1,164 @@
...
+  color: #00acdc;
+}
+
+a:not(:last-child) {
+  margin-right: 20px;
+}
+
+.is-active::after {
+  content: "";
+  width: 37px;
```
gestap: this class must be adaptive.
If you add this class to elements that are larger, what we will see?

src/style.css
```diff
@@ -0,0 +1,164 @@
+body {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+header {
+  width: 1200px;
```
gestap: no need to use @media. Just write `max-width: 1200px`

src/style.css
```diff
@@ -0,0 +1,164 @@
...
+    width: 1024px;
+  }
+}
+
+main {
+  box-sizing: border-box;
+  display: flex;
+  flex-wrap: wrap;
+  width: 1024px;
+  height: 853px;
```
gestap: Why did you add fixed value? 
If we want to add more card, what will be?

src/style.css
```diff
@@ -0,0 +1,164 @@
...
+    
+h3 {
+  margin: 0;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 15px;
+}
+  
+p {
+  margin: 6px 0 0;
```
gestap: use `margin-top` if you need only one side

PR #21, opened 2019-05-03
not ready, need help from TA

valeriitovstyk:


src/index.html
```diff
@@ -13,6 +13,202 @@
   <link rel="stylesheet" href="./style.css">
```
gestap: You forgot about link for Font

src/index.html
```diff
@@ -13,6 +13,202 @@
...
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star.svg" alt="star"></div>
+        </div>
+        <div class="reviews">Отзывов: 5</div>
+      </div>
+      <div class="price">
+        <div class="price_title">Цена:</div>
+        <div class="price_number">69 999 грн</div>
```
gestap: I think better to use `p` or `span`

src/index.html
```diff
@@ -13,6 +13,202 @@
...
+      <div class="item_code">Код товара: 195434</div>
+      <div class="rank">
+        <div class="stars">
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star-active.svg" alt="star"></div>
+          <div class="star"><img src="images/star.svg" alt="star"></div>
+        </div>
+        <div class="reviews">Отзывов: 5</div>
```
gestap: I think better to use `span`

src/style.css
```diff
@@ -0,0 +1,168 @@
...
+li {
+  margin-left: 20px;
+  display: flex;
+}
+
+li:first-child {
+  margin: 0;
+}
+
+a {
```
gestap: better to add `a` class

src/style.css
```diff
@@ -0,0 +1,168 @@
...
+  font-family: "Roboto", sans-serif;
```
gestap: `color` and `font-family` are inherited.
Better to put them into `body` and you don't need to duplicate them.

src/style.css
```diff
@@ -0,0 +1,168 @@
...
+.button_text:hover {
+  background-color: #fff;
+  color: #00acdc;
+}
+
+main {
+  margin: 50px 40px 40px;
+}
+
+.main {
```
gestap: Bad name for a class.
In your case, i've recommended something like `container`.

src/style.css
```diff
@@ -0,0 +1,168 @@
...
+
+.price {
+  padding: 21px 15px 0;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+}
+
+.price_title {
+  font-size: 11px;
```
gestap: `font-size` also inherited.
Add it to `body`.

PR #23, opened 2019-05-05
solution

ilovepinkpony1:
font weight in layout and reference is different, and i dont know what to do

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,180 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  color: #000;
+  font-family: "Roboto", sans-serif;
+}
+  
+.card_wrap {
+  background: #fff;
+  border: 1px solid #f3f3f3f3;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,180 @@
...
+  font-style: normal;
+  font-weight: 500;
+  font-size: 12px;
+  line-height: 15px;
+  margin-bottom: 5px;
+}
+
+.card_product_code {
+  font-style: normal;
+  font-weight: normal;
```
gestap: ` font-style: normal;
  font-weight: normal;`
also inherited.

src/style.css
```diff
@@ -0,0 +1,180 @@
...
+  justify-content: center;
+}
+  
+.card_img img {
+  width: 100%;
+  height: auto;
+}
+
+.card_product_description {
+  font-style: normal;
```
gestap: !

PR #108, opened 2019-07-27
Solution

marchukmiroslava88:


src/index.html
```diff
@@ -9,11 +9,275 @@
...
+  <title>MOYO</title>
+  <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" rel="stylesheet">
+  <link href="https://cdnjs.cloudflare.com/ajax/libs/meyer-reset/2.0/reset.min.css" rel="stylesheet">
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <header class="header">
+    <a class="header__logo" href="/">
+      <img src="images/logo.png" alt="">
```
irenhh: add something like 'moyo logo' to alt attribute

src/index.html
```diff
@@ -9,11 +9,275 @@
...
+  </header>
+  <main>
+    <div class="container">
+      <div class="product__list">
+        <div class="product__list-item">
+          <div class="product" data-qa="card">
+            <a href="#" class="product__photo">
+              <img src="images/imac.jpeg" alt="product-photo" title="Моноблок APPLE A1419 iMac 27 Retina 5K (MNED2UA/A)">
+            </a>
+            <a href="#" class="product__name">
```
irenhh: better to wrap it in heading tag

marchukmiroslava88: Please explain the meaning of your statement more fully. Wrap picture in h tag? I do not have h1 on the page, according to the seo rules I cannot avoid skipping heading levels. There is no title on the page, I don’t see where h1 can be added.

YuriiHoliuk: I think you are right

PR #144, opened 2019-09-03
Solution

FelineKing:
https://felineking.github.io/layout_catalog
https://felineking.github.io/layout_catalog/report/html_report/


FelineKing:
Even don't think about it, TREVOR!

PR #492, opened 2020-07-13
Flex

rmnkk:
- [DEMO LINK](https://rmnkk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://rmnkk.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,17 +11,21 @@
       <a href="#" class="header__logo">
         <img src="images/logo.png" alt="">
       </a>
+
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item">
             <a class="nav__link nav__link_active is-active" href="#">Apple</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Samsung</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Смартфоны и телефоны</a>
           </li>
+
           <li class="nav__item">
             <a
               class="nav__link"
@@ -29,15 +33,19 @@
               href="">Ноутбуки и компьютеры
             </a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Гаджеты</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Планшеты</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Фото</a>
           </li>
+
           <li class="nav__item">
             <a class="nav__link" href="#">Видео</a>
           </li>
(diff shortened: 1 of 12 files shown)


src/index.html
```diff
@@ -7,6 +7,352 @@
...
+          <li class="nav__item">
+            <a class="nav__link" href="#">Видео</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="catalog">
+        <div class="card catalog__card" data-qa="card">
```
irenhh, line 58:
```suggestion
        <div class="catalog__card card" data-qa="card">
```
usually it's written like this

src/styles/_reset.scss
```diff
@@ -0,0 +1,11 @@
+html,
+h2,
+body,
+ul,
+a {
+  font-family: $font-stack;
```
irenhh, line 6: assign font family to body only

irenhh: you don't need all those styles to the elements you've written there

src/styles/card.scss
```diff
@@ -0,0 +1,77 @@
+.catalog__card {
+  margin: 0 $card-horizontal-margin 50px;
+}
+
+.card {
+  width: $card-width;
+  border: 1px solid #f3f3f3;
+  box-sizing: border-box;
```
irenhh, line 8: this should be assigned to all elements with * selector

src/styles/card.scss
```diff
@@ -0,0 +1,77 @@
...
+  &__price-text {
+    font-size: $font-small;
+  }
+
+  &__value {
+    font-weight: 700;
+    font-size: $font-big;
+  }
+
+  &__button {
```
irenhh, line 59: why `a` tag is named as button?

src/index.html
rmnkk: Got it.

src/styles/_reset.scss
rmnkk: The reason why I added all of these tags here are next: 
For example, if I would delete h2 tag from here, text inside h2 tag will have default size and I would need to add font-size to the class which customizes h2 tag.
Also, I reset text-decoration for the "a" tag, and list-style for the ul, for not to write inside their classes.
The body tag is here for the margin: 0;
Html tag for font-family and font-size;

src/styles/card.scss
rmnkk: Okay. Fixed

src/styles/card.scss
rmnkk: I think it was previously in layout_product_cards task required to style button using an "a" tag.

src/styles/_reset.scss
irenhh: if you need only margins for body, then write them separately with body, because body may not need other styles

don't assign just for sure, use the styles you really need at this time

src/styles/card.scss
irenhh: then name it as link, or buy-link

rmnkk: Got it.

src/styles/_reset.scss
rmnkk: Also need margin 0 for h2 tag. So, decided to assign margin 0 both for body and h2.

PR #496, opened 2020-07-13
Catalog [Не грузятся референсы]

VenomSRT:
- [DEMO LINK](https://VenomSRT.github.io/layout_catalog/)
- [TEST REPORT LINK](https://VenomSRT.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -7,6 +7,382 @@
...
+
+      <nav class="nav">
+        <a href="#" class="nav__link nav__link--inner-style nav__link--is-active is-active">Apple</a>
+        <a href="#" class="nav__link nav__link--inner-style">Samsung</a>
+        <a href="#" class="nav__link nav__link--inner-style">Смартфоны и телефоны</a>
+        <a
+          data-qa="nav-hover"
+          href="#"
+          class="nav__link nav__link--inner-style"
+
```
irenhh, line 29: remove this

src/styles/_normalize.scss
```diff
@@ -0,0 +1,25 @@
+html {
+  font-family: $page-font;
+  text-decoration: none;
+  font-size: 12px;
+}
+
+body {
+  margin: 0;
+
+  * {
```
irenhh, line 10: why is it inside body?

src/styles/product_card.scss
```diff
@@ -0,0 +1,88 @@
...
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    margin: 0 0 45px;
+  }
+
+  &__info {
+    width: 170px;
+    flex-wrap: wrap;
```
irenhh, line 18: use flex-direction instead of this -> this will allow you to remove width 100% everywhere

src/index.html
```diff
@@ -7,6 +7,382 @@
...
+          <img
+            src="./images/logo.png"
+            alt="moyo logo"
+            class="header__site-logo"
+          >
+        </a>
+      </div>
+
+      <nav class="nav">
+        <a href="#" class="nav__link nav__link--inner-style nav__link--is-active is-active">Apple</a>
```
irenhh, line 22: why did you create extra class `nav__link--inner-style` for each link?

src/styles/_normalize.scss
VenomSRT: > why is it inside body?

```
  * {
     display: flex;
  }
```
makes <title> visible on page if rule isn't nested in body

src/index.html
VenomSRT: > why did you create extra class `nav__link--inner-style` for each link?

Misha said, that rules which describes positioning in parent block, should be separated from "internal" rules, so I made "inner" class. May be I misunderstood this.

src/styles/_normalize.scss
```diff
@@ -1,20 +1,24 @@
...
+}
 
-  * {
-    display: flex;
-    box-sizing: border-box;
-    margin: 0;
-  }
+body * {
+  display: flex;
 }
```
irenhh: don't do this, assign display value to each necessary element separately

src/index.html
irenhh: in your case the rules can be written in the same class

PR #591, opened 2020-11-10
Catalog was added and stylized

AlexysUa:
- [DEMO LINK](https://alexysua.github.io/layout_catalog/)
- [TEST REPORT LINK](https://alexysua.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm ci
+    - run: npm install
     - run: npm start & sleep 5 && npm test
+    - name: Upload HTML report(backstop data)
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: report
+        path: backstop_data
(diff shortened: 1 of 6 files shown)


Alexandr-Petrenko (approved):
Good job

src/styles/base/_reset.scss
```diff
@@ -0,0 +1,5 @@
+* {
+  box-sizing: border-box;
+  padding: 0;
+  margin: 0;
+}
```
aionova, line 5: Не задавай марджины и паддинги для всех элементов. Задавай только там, где они нужны

src/styles/components/_catalog.scss
```diff
@@ -0,0 +1,7 @@
+.catalog {
+  max-width: 992px;
```
aionova, line 2: Don't write magic numbers. It's always better to write the way you calculated this number with variables 
https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

aionova (approved):
Awesome! :fire:

src/styles/components/_catalog.scss
AlexysUa: Thanks for review! I'll fix it now.

src/styles/base/_reset.scss
AlexysUa: But 
https://css-tricks.com/reset-all-margins-padding/

PR #657, opened 2020-11-26
Grid Catalog

paramonovfrontend:
- [DEMO LINK](https://paramonovfrontend.github.io/layout_catalog/)
- [TEST REPORT LINK](https://paramonovfrontend.github.io/layout_catalog/report/html_report/)


5Mountains (approved):
Nice grid approach implementation! 👌
However, from my perspective, it might be more appropriate to apply a `grid` method only to the menu inside the `nav` rather than to the whole `header`.

vkryvytskyy (approved):

PR #758, opened 2021-04-01
Solution added

TimurHrynenko:
- [DEMO LINK](https://TimutHrynenko.github.io/layout_catalog/)
- [TEST REPORT LINK](https://TimutHrynenko.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,85 @@
+.card {
+  max-width: $card-max-width;
+  padding: 29px 14px 14px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &__img {
+    display: block;
+    margin: auto;
+    max-width: 160px;
+    min-height: 134px;
+    margin-bottom: 45px;
+  }
+
+  &__title {
+    margin-bottom: 6px;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 15px;
+  }
+  
+  &__code {
+    margin-bottom: 15px;
+    @include card-small-text;
+  }
+
+  &__rating {
+    margin-bottom: 20px;
+    @include classic-flex;
+  }
+
+  &__grade {
+    @include classic-flex;
+  }
+
+  &__star {
+    background: center / cover no-repeat url("../images/star.svg");
+    min-width: 15px;
+    min-height: 15px;
+    margin-right: 5px;
+
+    &_active {
+      background: center / contain no-repeat url("../images/star-active.svg");
+    }
+  }
+
+  &__pricing {
+    margin-bottom: 20px;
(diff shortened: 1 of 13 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,18 @@
...
+@import "./blocks/cards";
+@import "./blocks/card";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
+  font-family: Roboto, Arial, sans-serif;
```
victor-arutiunov, line 15: main font of page sets in html tag

TimurHrynenko: I think its can be done also in your way, but still all the content of the page is inside `body ` as well, so there not much difference, i suppose. 

Its only metters for property `font-size` if i or you will use rem or em, but i think they not very usfull, because the designer of the page not very often make some proportion for fonts)

vsm2k17 (changes requested):
лінки поламані

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,85 @@
...
+  padding: 29px 14px 14px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &__img {
+    display: block;
+    margin: auto;
+    max-width: 160px;
+    min-height: 134px;
+    margin-bottom: 45px;
```
aionova, line 12: margin-bottom is supposed to be less than 45px

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,18 @@
+.cards {
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+  max-width: getCardContainerWidth($card-max-width, $card-margin-size, 4);
+  margin: auto;
```
aionova, line 6: remove margins from BEM-block

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,49 @@
+.header {
+  padding: 0 50px;
+  @include classic-flex;
+  box-shadow: 0 2px 4px 0 rgba(0,0,0,0.05);
+}
+
+.nav {
```
aionova, line 7: It's better to create a separate file for each BEM-block

aionova (changes requested):
Please fix as close to figma as possible.

src/styles/blocks/card.scss
TimurHrynenko: But on the mocup there is 45px gap from picture to title of the card. Why should it be less?

aionova: you're right, my bad

aionova (approved):
Good job!

PR #772, opened 2021-04-02
Solution

bohdanklius:
- [DEMO LINK](https://bohdanklius.github.io/layout_catalog/)
- [TEST REPORT LINK](https://bohdanklius.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -5,8 +5,327 @@
...
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="logo">
+        <a href="#"><img src="./images/logo.png" alt="MOYO"></a>
```
Allrightman, line 16: As an alternative, you can use without div, because <a> is already wrapper

src/styles/_reset.scss
```diff
@@ -0,0 +1,88 @@
...
+mark,
+audio,
+video {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  font: inherit;
+  font-size: 100%;
+  vertical-align: baseline;
+}
```
Allrightman, line 88: I'm not sure, maybe this is normal practice to use prepared reset file, but you haven't so many tags.
Maybe you can use just some of them, or make reset: all, display: revert?
Mentor, help us please)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
...
+
+  &__name {
+    margin: 0 0 6px;
+    font-weight: $fw-semibold;
+    font-size: $fs-md;
+    line-height: 15px;
+  }
+
+  &__code {
+    margin: 0 0 15px;
```
Allrightman, line 25: there is information in checklist, that we should use margin/padding-[direction], if we have only one

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  &__rating {
+    width: 100%;
+    margin: 0 0 20px;
+    display: flex;
+    justify-content: space-between;
+  }
+
+  &__stars .stars__star:nth-child(-n+4) {
```
Allrightman, line 38: this should belong to stars block

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
...
+    background: url("../images/star-active.svg") no-repeat 50%;
+  }
+
+  &__feedback {
+    margin: 0;
```
Allrightman, line 43: you already reset margins in reset.scss?

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,48 @@
+.nav {
+  display: flex;
+  align-items: center;
+
+  &__list {
+    display: flex;
+    padding: 0;
+    margin: 0;
```
Allrightman, line 8: reset.scss?

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,10 @@
+.stars {
+  display: flex;
+  height: 15px;
+
+  &__star {
+    width: 15px;
+    margin-right: 4px;
+    background: url("../images/star.svg") no-repeat center;
+  }
+}
```
Allrightman, line 10: Misha mentioned that we have to make block stars, so it's worth adding to this block everything related to stars

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,92 @@
...
+    display: flex;
+    align-items: center;
+    font-size: $fs-sm;
+    line-height: 13px;
+  }
+
+  &__price {
+    width: 100%;
+    margin-bottom: 20px;
```
Allrightman, line 52: I'm not quite sure, but maybe we should use paddings instead of margins inside element (BEM)

Allrightman (approved):
Great, I'm not so lucky with tests))
Optional: you can use flex-shrink: 0; because text in nav goes to the column at width 937px

src/index.html
```diff
@@ -5,8 +5,327 @@
...
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="main">
+      <section class="main__catalog catalog">
+        <div class="catalog__card card" data-qa="card">
+          <img class="card__image" src="./images/imac.jpeg"
+            alt="APPLE iMac Retina 5K">
```
aliceblunt8, line 71:
```suggestion
          <img
            class="card__image"
            src="./images/imac.jpeg"
            alt="APPLE iMac Retina 5K"
          >
```

src/styles/_reset.scss
aliceblunt8: no strong opinion, bu I also think it's better to reset only those tags which you personally need in your project

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,85 @@
...
+  &__rating {
+    width: 100%;
+    margin-bottom: 20px;
+    display: flex;
+    justify-content: space-between;
+  }
+
+  &__feedback {
+    display: flex;
+    align-items: center;
```
aliceblunt8, line 40: why do you need these properties here, on flex child? just add align-items: center to the flex parent container

src/styles/blocks/card.scss
aliceblunt8: I think it's perfectly fine to use margins inside block (it's not the same as using margins outside the block - this is not good)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,17 @@
...
+
+  &__catalog {
+    max-width: 1024px;
+    display: flex;
+    justify-content: space-between;
+    flex-wrap: wrap;
+    padding: 0 40px;
+    margin-top: 50px;
+  }
+}
```
aliceblunt8, line 13: move to mix (main__catalog) only those properties which are related to outer spacing and positioning and move all other styles to catalogue itself

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,47 @@
+.nav {
+  display: flex;
+  align-items: center;
+
+  &__list {
+    display: flex;
+  }
```
aliceblunt8, line 7: why do you need display: flex on both nav and nav_list? leave flex just on list

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  &__item {
+    list-style: none;
```
aliceblunt8, line 10: add to reset file

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,47 @@
...
+  &__link {
+    height: 60px;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    color: $nav-text-color;
+    font-weight: $fw-semibold;
+    font-size: $fs-md;
+    line-height: 14px;
+    text-decoration: none;
```
aliceblunt8, line 23: move to reset file

aliceblunt8: text-decoration: none;

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  &__item {
+    list-style: none;
+    margin-left: 20px;
+  }
+
+  &__link {
+    height: 60px;
+    display: flex;
+    justify-content: space-between;
```
aliceblunt8, line 17: you use flex for centring, no need for justify-content: space-between;

aliceblunt8 (changes requested):

aliceblunt8 (approved):
please give attention to the comments

PR #778, opened 2021-04-03
Beautiful catalog

Allrightman:
[DEMO LINK](https://Allrightman.github.io/layout_catalog/)
[TEST REPORT LINK](https://Allrightman.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,86 @@
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+
+  margin: 0 48px 50px 0;
+  padding: 29px 14px 14px;
+  width: 200px;
+
+  border: 1px solid #f3f3f3;
+  box-shadow: 0 2px 5px rgba(0, 0, 0, 0.07);
+  border-radius: 5px;
+
+  &:nth-child(4n) {
+    margin-right: 0;
+  }
+
+  &__img {
+    align-self: center;
+
+    @include size(160px, 134px);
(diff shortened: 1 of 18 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+
+  margin: 0 48px 50px 0;
```
Allrightman, line 6: I used margins because gap on wrapper doesn't work with tests (they don't see gaps)

bohdanklius (approved):
Молодець! Начебто все ок)

src/styles/blocks/card.scss
aionova: You can add another class to each card 'catalog__card' and give margins to it

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,9 @@
+.catalog {
+  // gap: 48px;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  margin: auto;
```
aionova, line 7: Here too. You can add another class main__catalog and include margins there

aionova (changes requested):
Almost done)

src/styles/blocks/catalog.scss
Allrightman: really, I can implement mixes, thanks!

aionova (approved):
Awesome!

PR #818, opened 2021-04-06
Smooth changes

cheytac200:
- [DEMO LINK](https://cheytac200.github.io/layout_catalog/)


src/styles/blocks/main.scss
```diff
@@ -0,0 +1,125 @@
...
+    margin-right: 0;
+  }
+
+  &__image {
+    width: 160px;
+    height: 134px;
+
+    &:hover {
+      transform: translateY(-14px) scale(1.21);
+      transition: 400ms;
```
aionova, line 43:
```suggestion
      transition: transform 400ms, box-shadow 400ms, border-radius 400ms;
```
Include properties that you use transition for, and move transition to &__image from &:hover.

aionova (approved):
The card itself is not supposed to get bigger on hover. Only the image inside the card.

aionova (approved):
Good job! 
According to mockup, image padding-top inside the card needs to be bigger. But it still looks good

PR #1134, opened 2021-10-18
catalog created

Val-space:
[DEMO LINK](https://Val-space.github.io/layout_catalog/)
[TEST REPORT LINK](https://Val-space.github.io/layout_catalog/report/html_report/)
can't pass the test for Nav_tag_1024px.png

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,4 +1,7 @@
 .card {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
   width: $card-width;
   border: 1px solid #f3f3f3;
   border-radius: 5px;
@@ -77,4 +80,3 @@
     }
   }
 }
-


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,55 @@
+@import '../utils/mixins';
+
+.header {
+ @include flex-base;
+  padding: 0 50px;
+  box-shadow: 0px 2px 4px 0px #0000000D;
+
+  &__logo {
+    padding: 10px 0;
+  }
```
dmtrhrytsak, line 10: It doesn't do anything :D

```suggestion
```

dmtrhrytsak: You already have flex for centering)

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,55 @@
...
+  &__list {
+    display: flex;
+  }
+
+  &__li:not(:first-child) {
+    margin-left: 20px;
+  }
+
+  &__link {
+    display: block;
```
dmtrhrytsak, line 29: **display: inline-block** would suffice! It lets you define top/bottom paddings, then why use such a radical approach display: block? :d

```suggestion
    display: inline-block;
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+@import '../utils/mixins';
+
+.card {
+  width: $card-width;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  color: #060b35;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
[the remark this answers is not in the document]
ihor-jpeg: why not?

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,82 @@
...
+  &__price-sum {
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 18px;
+  }
+
+  &__button {
+    display: inline-block;
+    background-color: #00acdc;
+    width: 100%;
```
ihor-jpeg, line 64: items inside a flex container will automatically have the display property set to block. and the block elements have 100% width by default.
for that reason, display and width are unnecessary here

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,10 @@
+.catalog {
+  padding: 50px 0 73px;
+  
+  &__cards {
+    display: flex;
+    flex-wrap: wrap;
+    margin: 0 auto;
+    max-width: ($card-width * 4) + ($card-gap * 3);
+  }
+}
```
ihor-jpeg, line 10: add a new line at the end of each file

src/styles/blocks/reset.scss
```diff
@@ -0,0 +1,20 @@
...
+}
+
+ul {
+  padding: 0;
+  margin: 0;
+}
+
+li {
+  list-style: none;
+}
```
ihor-jpeg, line 20: empty line at the end

src/styles/blocks/card.scss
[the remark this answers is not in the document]
dmtrhrytsak: > why not?

They stack onto each other in column fashion without flex, and distance between block isn't equal (we specify it with margins), so justify-content: space-between is useless. I might be wrong, it just seems to me so

PR #1150, opened 2021-10-20
Catalog

MagicBox15:
- [DEMO LINK](https://<MagicBox15>.github.io/layout_catalog/)
- [TEST REPORT LINK](https://<MagicBox15>.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Frontend practice with catalog page
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<MagicBox15>.github.io/layout_catalog/)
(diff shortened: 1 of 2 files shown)


src/styles/blocks/card/card.scss
```diff
@@ -0,0 +1,98 @@
...
+    line-height: $line-height-big;
+    font-weight: 500;
+    color: $color-main;
+  }
+
+  &__product-code {
+    padding-top: 4px;
+    font-size: $font-size-small;
+    line-height: $line-height-small;
+    font-weight: $font-weight-normal;
```
ihor-jpeg, line 27: no need to set this explicitly. normal is normal even without setting it to normal.

ihor-jpeg (approved):
Good work, but write less code. have a look at the comments

PR #1152, opened 2021-10-20
Catalog was created

Vitalii-Lytvyn01:
[DEMO LINK](https://Vitalii-Lytvyn01.github.io/layout_catalog/)
[TEST REPORT LINK](https://Vitalii-Lytvyn01.github.io/layout_catalog/report/html_report/)
Some tests didn't pass, but it looks the same as reference and should work as intended.


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,119 @@
...
+
+  &:nth-child(4n) {
+    margin-right: 0;
+  }
+
+  &:nth-child(n+5) {
+    margin-top: 46px;
+  }
+
+  .card__image {
```
alex-kostomarov, line 21: .card elements should be inside BEM block and used "&"

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,119 @@
...
+    &:hover {
+      color: #00acdc;
+      background: #fff;
+      border: 1px solid #00acdc;
+    }
+  }
+}
+
+/* #region Stars */
+.stars {
```
alex-kostomarov, line 95: stars is a separate block, so it should be in a separate file

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,8 @@
+.catalog {
+  display: flex;
+  width: $catalog-width;
+  margin: 0 auto;
+  padding: 50px $catalog-padding 73px;
```
alex-kostomarov, line 5: Maybe side paddings is redundant. Block "catalog" horizontally centered because of "margin: 0 auto;". So if paddings will be removed block "catalog" will stay in the same position, just its width become smaller by 80px (two 40px paddings).

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,57 @@
...
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  .header__logo {
+    height: 40px;
+  }
+}
+
+.nav__list {
```
alex-kostomarov, line 15: according to the current naming ".nav" is a separate block and have to be in separate file

src/styles/blocks/catalog.scss
alex-kostomarov: Ok, about paddings that was just suggestion and you don't need to fix it

PR #1158, opened 2021-10-20
catalog

TatianaMorozova04:
- [DEMO LINK](https://TatianaMorozova04.github.io/layout_catalog/)
- [TEST REPORT LINK](https://TatianaMorozova04.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/cards.scss
+++ b/src/styles/blocks/cards.scss
@@ -1,8 +1,14 @@
-.cards {
+.cards__element {
+  width: $card-width;
   display: flex;
-  flex-wrap: wrap;
-  box-sizing: content-box;
-  width: $page-width;
-  padding: 50px 40px 72px;
-  margin: 0 auto;
+  flex-direction: column;
+  padding: 32px 16px 16px;
+
+  &:not(:nth-child(4n)) {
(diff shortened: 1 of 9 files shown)


src/styles/blocks/rating.scss
```diff
@@ -0,0 +1,36 @@
+.rating {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding-bottom: 24px;
+
+  &__text {
+    font-family: $page-font;
```
Val-space, line 8: font weight 400 is default, and you use the same font everywhere, add it once to the html tag

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,8 @@
+.cards {
+  width: $page-width;
```
Val-space, line 2: It is said in the homework that page width is always >= 1024px, but  I don't think it means the card width should be the same. You can create a variable that will count the cards' max-width. (multiply card width by 4 + gaps between the cards), so there always will be 4 cards in a row. And you won't need to use 'padding: 0 40px;' here ;

PR #1379, opened 2022-02-21
master,develop were merged

HelenPetrovska:
https://helenpetrovska.github.io/layout_catalog/
https://helenpetrovska.github.io/layout_catalog/report/html_report/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -107,7 +107,7 @@ tells the user that there is no warranty for the work (except to the
 extent that warranties are provided), that licensees may convey the
 work under this License, and how to view a copy of this License.  If
 the interface presents a list of user commands or options, such as a
-menu, a prominent item in the list meets this criterion.
+nav, a prominent item in the list meets this criterion.
 
   1. Source Code.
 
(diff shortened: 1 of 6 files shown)


src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+@mixin font($size: 14px, $color:$basic-color, $weight: 400, $lh: 1) {
+  font-size: $size;
+  color: $color;
+  font-weight: $weight;
+  line-height: $lh;
```
strukk311, line 22: Your code is so good, really))
You shouldn't use variable when you need to use that value just once ($font-size / $font-weight).

strukk311 (changes requested):

src/styles/utils/_mixins.scss
HelenPetrovska: Thanks) This mixin is general for all font styles, so I like it. Because I can write all font styles in one line)

PR #1434, opened 2022-03-16
catalog

DikanevYura:


src/styles/main.scss
```diff
@@ -1,3 +1,245 @@
...
+  margin-right: 0;
+}
+
+.header-nav__link {
+  text: {
+    decoration: none;
+    transform: uppercase;
+  }
+  font: {
+    family: $page-font;
```
Bohdan-Maksymiuk, line 55: No need. This styles are inherited from parent. You can write them in one parent.

src/styles/main.scss
```diff
@@ -1,3 +1,245 @@
...
+  background-color: $color-blue;
+  width: 100%;
+  left: 0;
+  bottom: 0;
+}
+
+.catalog {
+  display: flex;
+  flex-wrap: wrap;
+  margin: auto;
```
Bohdan-Maksymiuk, line 101: justify-content: center;

Bohdan-Maksymiuk (changes requested):
cards container should have fixed paddings (see 1024px example), bottom the same as top
Add horizontal padding
Screenshot 2022-03-16 at 15 07 13

src/styles/main.scss
DikanevYura: I hope I fixed everything)

aliceblunt8 (approved):

PR #1466, opened 2022-04-06
Catalog first try

enrikadej:


src/index.html
```diff
@@ -4,9 +4,362 @@
...
+        <div class="card__review">
+          <div class="stars stars--4">
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+
+          <span class="card__amount-reviews">Отзывов: 5</span>
```
dellachaise, line 96: Please use tag `p` instead of `span` in this case

src/index.html
```diff
@@ -4,9 +4,362 @@
...
+        </div>
+
+        <div class="card__price">
+          <span>Цена:</span>
+          <span class="card__price-amount">69 999 грн</span>
+        </div>
+        <a
+          class="card__link"
+          href="/"
```
dellachaise, line 105: You can use `#` instead of `/` if you haven't adress

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  flex-direction: column;
+  padding: 0 16px 16px;
+
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  box-sizing: border-box;
+
+  &__imac-photo {
+    display: block;
+    width: 100%;
```
dellachaise, line 13: try to add width as in mockup< in this case you can delete padding: 0 3px;, I think

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  &__review {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: center;
+    align-content: center;
+
+    width: #{$weight-card};
```
dellachaise, line 46: You can write this variable without #{}, it's redundant. Please delete it in other places.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
...
+    width: 40px;
+    height: 40px;
+  }
+}
+
+.logo {
+  display: block;
+}
+
+.nav {
```
dellachaise, line 20: please create a new file for nav block `nav.scss` and remove there nav block

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,21 @@
...
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: center;
+
+    width: 16px;
+    height: 16px;
+    margin-right: 4px;
+  }
+
+  @for $index from 1 through 5 {
```
dellachaise, line 16: @for loop should be nested to &__star element

dellachaise (changes requested):

src/styles/blocks/stars.scss
enrikadej: that`s okay, because I use `&--#{$index} > &__star:nth-child(-n + #{$index})` , so I don`t need to put in, I think

src/styles/blocks/card.scss
enrikadej: Anya helped me to pass some tests with these paddings, so let it be)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,23 @@
+
+.catalog {
+  display: flex;
+  flex-wrap: wrap;
+  padding: $paddingY $paddingX;
+  width: (4 * $cardWrapWidth) + (3 * $gapX) + (2 * $paddingY);
+  margin: 0 auto;
```
Anna-Tkachenko, line 7: You can't set margins for BEM block

Anna-Tkachenko (approved):
Good job! 👍

PR #1605, opened 2022-06-03
npm test doesnt work

serhii-yunnikov:
https://serhii-yunnikov.github.io/layout_catalog/
https://serhii-yunnikov.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/blocks/container.scss
+++ b/src/styles/blocks/container.scss
@@ -1,0 +1,19 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  margin: auto;
+  width: 4 * $cardWidth + 3 * $gap;
+  padding-top: 51px;
+
+  & > :nth-child(n) {
+    margin-right: $gap;
+  }
+
+  & > :nth-child(4n) {
+    margin-right: 0;
+  }
+
+  & > :nth-child(n + 5) {
+    margin-top: 48px;
+  }
+}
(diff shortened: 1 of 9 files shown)


src/styles/blocks/container.scss
```diff
@@ -0,0 +1,19 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  margin: auto;
+  width: 4 * $cardWidth + 3 * $gap;
+  padding-top: 51px;
+
+  & > :nth-child(n) {
+    margin-right: $gap;
+  }
```
Bohdan-Maksymiuk, line 10:
```suggestion
  &__card:nth-child(n) {
    margin-right: $gap;
  }
```
it would be better

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,19 @@
...
+
+  & > :nth-child(4n) {
+    margin-right: 0;
+  }
+
+  & > :nth-child(n + 5) {
+    margin-top: 48px;
+  }
```
Bohdan-Maksymiuk, line 18: same

Bohdan-Maksymiuk (approved):
Good job, fix the comment 👍

src/styles/blocks/container.scss
serhii-yunnikov: container and card are  two different BEM blocks(card is not element of container). &__card :nth-child(n) doesnt work in my code. when I set just  & :nth-child(n), styles implement to all levels children of card class.  Similar way:  & .card :nth-child(n) -the same case, all levels children accept these styles settings.

PR #1626, opened 2022-06-09
catalog was created

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/layout_catalog/)
- [TEST REPORT LINK](https://1stIndominus.github.io/layout_catalog/report/html_report/)


.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
```
vkryvytskyy, line 2: image
It's better to use `flex-start` instead of `space-between`, so the structure will remain even if there are not enough cards

vkryvytskyy (approved):

PR #1627, opened 2022-06-09
catalog added

Yaroslav-Radchuk:
- [DEMO LINK](https://Yaroslav-Radchuk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Yaroslav-Radchuk.github.io/layout_catalog/report/html_report/)


.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
```
vkryvytskyy, line 2: image
It's better to use `flex-start`, so it the structure will remain if there are not enough cards on the page

vkryvytskyy (approved):

PR #1648, opened 2022-06-13
catalog task is done

katerinahradinarova:
- [DEMO LINK](https://katerinahradinarova.github.io/layout_catalog/)
- [TEST REPORT LINK](https://katerinahradinarova.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -4,9 +4,401 @@
...
+
+        <div class="card__prices">
+          <p class="card__price">Цена:</p>
+          <p class="card__amount">69 999 грн</p>
+        </div>
+
+        <a
+           href="#"
+           class="card__buy"
+           data-qa="hover"
```
Oleksandr-Pustovit, line 168: All the previous data-qa should be removed
```suggestion
```

src/styles/catalog.scss
```diff
@@ -0,0 +1,123 @@
+.catalog {
+  display: flex;
+  margin: 0 auto;
+  flex-wrap: wrap;
+  width: $catalog-width;
+  padding: 51px 0 0;
```
Oleksandr-Pustovit, line 6: bottom paddings should be the same as top (by the task)
```suggestion
  padding: 51px 0;
```

src/styles/catalog.scss
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.card {
```
Oleksandr-Pustovit, line 9: Card is a separate block according to BEM. It should be in its own file (src/styles/blocks/card.scss)

src/styles/catalog.scss
```diff
@@ -0,0 +1,123 @@
...
+  background: #fff;
+  width: 200px;
+  display: flex;
+  flex-direction: column;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  font-size: 10px;
```
Oleksandr-Pustovit, line 17: I think it was the cllass where we should begin using right property order
- [checklist](https://github.com/mate-academy/layout_catalog/blob/master/checklist.md) Number 12

I've seen other moments based on this rule further in code but I will only leave this comment for you to fix all issues.
```suggestion
  display: flex;
  flex-direction: column;
  width: 200px;
  padding: 32px 16px 16px;
  
  font-size: 10px;
  
  background: #fff;
  border: 1px solid #f3f3f3;
  border-radius: 5px;
```

src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
-body {
-  margin: 0;
-}
+@import "./variables.scss";
+@import "./header.scss";
+@import "./catalog.scss";
+@import "./reset.scss";
+@import "./star.scss";
```
Oleksandr-Pustovit, line 5: Import downloads the whole css file and applies all of its content to the final css file. It is important here to watch the element order.
For example here you will use reset.scss after catalog and header styles has been used.
```suggestion
@import "./variables.scss";
@import "./reset.scss";
@import "./header.scss";
@import "./catalog.scss";
@import "./star.scss";
```

src/styles/star.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin-right: 0;
+}
+
+.stars--1 :nth-child(-n+1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5) {
+  background-image: url(../images/star-active.svg);
+}
```
Oleksandr-Pustovit, line 25: I have the same aproach here, but I am sure mentors will ask to do it with some kind of SCSS function or loop or so.
```@for``` is good here 
anyway practice will never harm

**Review by Oleksandr-Pustovit — CHANGES_REQUESTED** [review]
Great one :)
Some comments are attached

File: src/styles/active-state.scss [hunk 7]
```diff
@@ -0,0 +1,14 @@
+.is-active {
```
Oleksandr-Pustovit, line 1: I think this classes should stay with "nav" block. We had to name this class "is-active" only in this task while according to BEM it should be named (nav__link--is-active). So technically this refers to "nav.scss" file.

Oleksandr-Pustovit (approved):
Well done!
One last thing I noticed is file tree you have. generally it should be separated into folders like this
Stlyles
|-blocks
| |-"block1".scss
| |-"block2".scss
| |-"block3".scss
|
|-utils
| |-"_variables".scss
| |-"_functions".scss
| |-"_mixins".scss
| |-"_extentions".scss
|
|-main.scss
|-reset.scss

src/styles/catalog.scss
```diff
@@ -0,0 +1,19 @@
+.catalog {
+  display: flex;
+  flex-wrap: wrap;
+  padding: 50px calc((100% - #{$catalog-width}) / 2);
```
Ihor-Karpyn, line 4: Better calculate (without calc-func) container's max-width. This solution it's not good 
```suggestion
  padding: 50px 0;
```

Ihor-Karpyn (approved):

PR #1702, opened 2022-06-26
Rudenko done

RudBeee:
- [DEMO LINK](https://rudbeee.github.io/layout_catalog/)
- [TEST REPORT LINK](https://rudbeee.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,392 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a class="header__logo" href="#">
+        <img src="images/logo.png" alt="Moyo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#">Apple</a>
+          </li>
+          <li class="nav__item" >
+            <a class="nav__link" href="#">Samsung</a>
+          </li>
+          <li class="nav__item">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -7,6 +7,392 @@
...
+          class="card">
+            <div class="card__image">
+              <img 
+                src="./images/imac.jpeg" 
+                alt="Imac" 
+                class="card__img">
+            </div>
+            <div class="card__title">
+              Моноблок APPLE A1419 iMac 27" Retina 5K (MNED2UA/A)
+            </div>
```
Bohdan-Maksymiuk, line 61: isn't it `h`?

src/index.html
```diff
@@ -7,6 +7,392 @@
...
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="main">
+      <div class="container">
+        <div class="container__card">
+          <div 
+          data-qa="card" 
+          class="card">
```
Bohdan-Maksymiuk, line 52:
```suggestion
          <div 
            data-qa="card" 
            class="card"
          >
```

Bohdan-Maksymiuk (changes requested):

you could put styles into separate files according to BEM blocks

src/index.html
RudBeee: yes it is h. But I don't know which p to set, depending on the context of the page.
h3?

Bohdan-Maksymiuk (approved):
Good 👍

PR #1731, opened 2022-07-13
 layout_catalog task complete

EhorSaprykin:


src/styles/main.scss
```diff
@@ -1,3 +1,232 @@
...
+$main-dark-color: #060b35;
+$main-blue-color: #00acdc;
+
+* {
+  padding: 0;
   margin: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.header {
```
ArtemTopolskyi, line 12: don't write all styles in single file, create separate file for every BEM block

ArtemTopolskyi (changes requested):

src/styles/main.scss
EhorSaprykin: done

EhorSaprykin: @ArtemTopolskyi check changes

PR #1773, opened 2022-07-19
Develop

Yaroslav-OS:
- [DEMO LINK](https://Yaroslav-OS.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Yaroslav-OS.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,107 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 198px;
+  height: 406px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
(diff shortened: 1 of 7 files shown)


oksana-logos-frontend (approved):
You often use width, height, font-size, font-weight for them you could create a @mixin, it would shorten your code, if we have some important size or main color, better create variable.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,106 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 198px;
+  height: 406px;
```
AlinaMatuschak, line 5: don't hard-code height, use margins

PR #1895, opened 2022-08-19
Develop

grafit-off:
- [DEMO LINK](https://grafit-off.github.io/layout_catalog/)
- [TEST REPORT LINK](https://grafit-off.github.io/layout_catalog/report/html_report/)

_Why i'm using hidden heading in section? Becouse according to documentation W3 sections and article should always have a heading, with very few exceptions, so according to the first rule in checklist, i'm added several heading._

### DIFF
--- a/src/styles/blocks/stars.scss
+++ b/src/styles/blocks/stars.scss
@@ -1,0 +1,22 @@
+.stars {
+  display: flex;
+
+  &__star {
+    width: 16px;
+    height: 16px;
+
+    background: url("../images/star.svg") center no-repeat;
+  }
+
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  &--5 &__star,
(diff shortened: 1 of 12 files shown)


src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,22 @@
+.stars {
+  display: flex;
+
+  &__star {
+    width: 16px;
+    height: 16px;
```
vmamchur, line 6: you could use mixin here, but I think it's not really important =)

vmamchur (approved):
🤯👍🏻

src/styles/blocks/stars.scss
grafit-off: Really, i'm forget about it, will change this soon. Thx

MaksimusCRS (approved):
GJ

PR #1896, opened 2022-08-19
solution

vmamchur:
- [DEMO LINK](https://vmamchur.github.io/layout_catalog/)
- [TEST REPORT LINK](https://vmamchur.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,12 +1,364 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Catalog</h1>
+  <body class="page__content">
+    <header class="header">
+      <a class="header__logo" href="#home">
+        <img src="images/logo.png" alt="moyo-logo">
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#laptops&computers"
+              class="nav__link"
+              data-qa="nav-hover"
+              >
+              laptops & computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#tablets" class="nav__link">tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#photo" class="nav__link">photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="content">
+      <div class="content__box">
+        <article class="card content__card" data-qa="card">
+          <img
+            class="card__image"
+            src="images/imac.jpeg"
+            alt="iMac-img"
+          >
+
+          <h2 class="card__title">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h2>
+
+          <p class="card__product-code">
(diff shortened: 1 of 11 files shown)


grafit-off (approved):
Good job 🔥. Left a few comments below.

src/index.html
```diff
@@ -1,12 +1,364 @@
...
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#laptops&computers"
+              class="nav__link"
+              data-qa="nav-hover"
+              >
```
OleksiiBlezniuk, line 34:
```suggestion
              class="nav__link"
              data-qa="nav-hover"
            >
```

src/index.html
```diff
@@ -1,12 +1,364 @@
...
+      </nav>
+    </header>
+
+    <main class="content">
+      <div class="content__box">
+        <article class="card content__card" data-qa="card">
+          <img
+            class="card__image"
+            src="images/imac.jpeg"
+            alt="iMac-img"
```
OleksiiBlezniuk, line 64: It's a bad practice to use `image` or `picture` (img even worse) in alt attribute
Alt attribute should contain image description, it's obvious that it's an image, and don't forget that screenreader used it too

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    font-size: 10px;
```
OleksiiBlezniuk, line 48: This value was used before, so let's create a variable

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+
+    margin-bottom: 16px;
+
+    &-text {
+      font-size: 12px;
+      line-height: 18px;
```
OleksiiBlezniuk, line 61: These value were used before, so let's create a variable

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+
+      color: $secondary-color;
+    }
+
+    &-value {
+      font-weight: 700;
```
OleksiiBlezniuk, line 67: This value was used before, so let's create a variable

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,14 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  padding: 0 50px;
```
OleksiiBlezniuk, line 6: This value was used before, so let's create a variable

OleksiiBlezniuk (changes requested):

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,50 @@
...
+    margin-right: 20px;
+  }
+
+  &__link {
+    display: inline-block;
+
+    text-decoration: none;
+    color: $accent-color;
+
+    font-weight: 500;
```
OleksiiBlezniuk, line 21: Looks like generic font-weight value, we can create variable for it

OleksiiBlezniuk (changes requested):
there should always be 4 cards in a row (not 2, 3 or 5)
Screenshot 2022-08-23 at 13 29 37

src/index.html
```diff
@@ -1,12 +1,364 @@
...
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="content">
+      <div class="content__box">
```
OleksiiBlezniuk, line 59: Do you really need this additional div?

src/styles/blocks/content.scss
```diff
@@ -0,0 +1,21 @@
+.content {
+  &__box {
+    margin: 0 auto;
+    padding: $page-padding 0;
+    max-width: $content-box-width;
+    min-width: $content-box-width;
+
+    display: flex;
+    flex-wrap: wrap;
+  }
```
OleksiiBlezniuk, line 10: You could apply this styles to `content` class (except margins, you need to create BEM element for margins) and avoid additional div here. Also it's strange that we set outer distance between element and parent container

OleksiiBlezniuk (approved):

src/index.html
vmamchur: idk why, but when I delete the additional div, this happens ⬇
image
everything looks fine for me, but the tests fail. mb I had to deploy failed tests, anyway visually everything was fine.

PR #1901, opened 2022-08-19
Catalog was created

anastasiia-nahaiska:
- [DEMO LINK](https://anastasiia-nahaiska.github.io/layout_catalog/)
- [TEST REPORT LINK](https://anastasiia-nahaiska.github.io/layout_catalog/report/html_report/)

I used em and rem for font-size in this code. I've read some pages on the Internet and some developers think that it's good, some don't.  How is better: px or em/rem?

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,86 @@
+.card {
+  width: $cardWidth;
+
+  border: 1px solid rgba(243, 243, 243, 1);
+  border-radius: 5px;
+
+  font-weight: 400px;
+  color: $lightTextColor;
+
+  padding: 32px 16px 16px;
+
+  &__image {
+    display: block;
+    width: 160px;
+    height: 134px;
+
+    margin: 0 3px 40px ;
+  }
+
+  &__name-of-product {
+    font-weight: 500;
+    font-size: 1.2em;
+    line-height: 18px;
+    color: $darkTextColor;
+
+    margin-bottom: 4px;
+  }
+
+  &__code-of-product {
+    margin-bottom: 16px;
+
+    line-height: 14px;
+  }
+
+  &__product-rating {
+    @extend %centringContent;
+
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    color: $darkTextColor;
+  }
+
+  &__price {
+    @extend %centringContent;
+
+    margin-bottom: 16px;
+  }
+
+  &__price-text {
+    font-size: 1.2em;
+    line-height: 18px;
+  }
+
+  &__price-value {
+    font-size: 1.6em;
+    line-height: 18px;
+    font-weight: 700;
+    color: $darkTextColor;
+  }
+
+  &__buy-button {
+    @extend %centringContent;
+
+    background-color: $styleColorMoyo;
+    border: 1px solid $styleColorMoyo;
+    border-radius: 5px;
+
+    padding: 10px 68px 11px;
+
+    color: #fff;
+    text-decoration: none;
+    font-size: 1.4em;
+    font-weight: 700;
+    line-height: 16px;
+    text-transform: uppercase;
+
+    transition: color 0.3s ease, background-color 0.3s ease;
+  }
+
+  &__buy-button:hover {
+    color: $styleColorMoyo;
+    background-color: #fff;
+  }
+}
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,0 +1,22 @@
+.catalog {
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+
+  padding: 50px 0;
+
+  max-width: $catalogWidth;
+  min-width: $catalogWidth;
+}
+
+.catalog__card {
+  margin: 0 48px 47px 0;
+}
+
+.catalog__card:nth-child(4n) {
+  margin-right: 0;
+}
+
+.catalog__card:nth-last-child(-n+4) {
+  margin-bottom: 0;
+}
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,0 +1,22 @@
+.header {
+  min-width: 1024px;
+
+  @extend %centringContent;
+
+  padding: 0 50px;
+
+  background-color: rgba(255, 255, 255, 1);
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+
+  font-size: 1.2rem;
+  color: rgba(0, 0, 0, 1);
+
+  &__logo-link {
+    display: flex;
+  }
+
+  &__logo {
+    width: 40px;
+    height: 40px;
+  }
+}
--- a/src/styles/blocks/nav.scss
+++ b/src/styles/blocks/nav.scss
@@ -1,0 +1,56 @@
+
+.nav {
+  &__list {
+    display: flex;
+
+    list-style: none;
+  }
+
+  &__item {
+    margin: 0 10px;
+  }
+
+  &__link {
+    display: flex;
+    text-decoration: none;
+    text-transform: uppercase;
+    font-weight: 500;
+    color: $darkTextColor;
+    transition: color 0.3s;
+
+    padding: 23px 0;
+  }
+
+  &__item:first-child {
+    margin-left: 0;
+  }
+
+  &__item:last-child {
+    margin-right: 0;
+  }
+
+  &__link:hover {
+    color: $styleColorMoyo;
+  }
+}
+
+.is-active {
+  position: relative;
+
+  color: $styleColorMoyo;
+}
+
+.is-active::after {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+
+  content: "";
+
+  background: $styleColorMoyo;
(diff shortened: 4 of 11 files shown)


ArtemHulkoDev (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+
+    color: #fff;
+    text-decoration: none;
+    font-size: 1.4em;
+    font-weight: 700;
+    line-height: 16px;
+    text-transform: uppercase;
+
+    transition: color 0.3s ease, background-color 0.3s ease;
+  }
```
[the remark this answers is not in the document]
anastasiia-nahaiska: In this case, will the transition apply to all styles? Maybe the mentor will tell me how to do it better because in the previous task, another student advised me to write specific styles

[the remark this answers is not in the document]
ArtemHulkoDev: yeah, it will work correctly, I have checked it already, but I agree with you, it is better to get the right answer from the mentor :)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,22 @@
+.catalog {
+  display: flex;
+  justify-content: center;
+  flex-wrap: wrap;
+
+  padding: 50px 0;
```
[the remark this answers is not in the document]
anastasiia-nahaiska: Oh, I forgot about this... Thanks!

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,22 @@
+.header {
+  min-width: 1024px;
```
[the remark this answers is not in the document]
anastasiia-nahaiska: Without min-width we have this problem if the width of a screen will be less than 1024px, so I think it necessary (of course if we haven't @media for a small screen)
image

src/styles/main.scss
```diff
@@ -1,3 +1,21 @@
...
 }
+
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+}
+
+.content__catalog {
+  margin: 0 auto;
+}
```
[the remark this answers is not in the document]
anastasiia-nahaiska: I decided to write in main.scss, because .content _catalog is the element of main. Overall I think this style must be in content.scss (file for semantic block main)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  max-width: $catalogWidth;
+  min-width: $catalogWidth;
```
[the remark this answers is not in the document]
ArtemHulkoDev: It seems we need to get a mentor here :) He/she will explain to us the right meaning of this one)

ArtemHulkoDev (approved):
everything is fine now, GJ :)

OleksiiBlezniuk (commented):
> I used em and rem for font-size in this code. I've read some pages on the Internet and some developers think that it's good, some don't. How is better: px or em/rem?

em/rem is cool, because if you'll change font-size in one place, other related values will automatically update

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
+.card {
+  width: $cardWidth;
+
+  border: 1px solid rgba(243, 243, 243, 1);
+  border-radius: 5px;
+
+  font-weight: 400px;
```
OleksiiBlezniuk, line 7: 1 - font-weight is a value without units, it's not working now
2 - it's a default value, so it can be removed

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+
+  &__price {
+    @extend %centringContent;
+
+    margin-bottom: 16px;
+  }
+
+  &__price-text {
+    font-size: 1.2em;
+    line-height: 18px;
```
OleksiiBlezniuk, line 53: You can also use em for line-height

src/styles/blocks/card.scss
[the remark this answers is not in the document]
OleksiiBlezniuk: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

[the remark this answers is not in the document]
OleksiiBlezniuk: @ArtemHulkoDev it works, but better not to do it 🙂 It's bad for page performance

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+    text-transform: uppercase;
+
+    transition: color 0.3s ease, background-color 0.3s ease;
+  }
+
+  &__buy-button:hover {
+    color: $styleColorMoyo;
+    background-color: #fff;
+  }
+}
```
OleksiiBlezniuk, line 86: use sass nesting for pseudoclasses

src/styles/blocks/catalog.scss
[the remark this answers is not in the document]
OleksiiBlezniuk: Everything is correct, it's required in task to always have 4 cards in a row, so we should set this width limits 🙂

src/styles/blocks/header.scss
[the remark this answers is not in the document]
OleksiiBlezniuk: min-width - good solution

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  @extend %centringContent;
+
+  padding: 0 50px;
+
+  background-color: rgba(255, 255, 255, 1);
```
OleksiiBlezniuk, line 8: Use variables for colors

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,22 @@
...
+  font-size: 1.2rem;
+  color: rgba(0, 0, 0, 1);
+
+  &__logo-link {
+    display: flex;
+  }
+
+  &__logo {
+    width: 40px;
+    height: 40px;
```
OleksiiBlezniuk, line 20: Let's create a mixin for it, we also need square for star elemen

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+.nav {
+  &__list {
+    display: flex;
+
+    list-style: none;
+  }
+
+  &__item {
+    margin: 0 10px;
+  }
```
OleksiiBlezniuk, line 11: Set margins for one direction, not for both. It's more clear and also you can remove it from one child, not for both first and last

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &__link {
+    display: flex;
+    text-decoration: none;
+    text-transform: uppercase;
+    font-weight: 500;
```
OleksiiBlezniuk, line 17: Looks like generic font-weight value, so let's create variable

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+    color: $darkTextColor;
+    transition: color 0.3s;
```
OleksiiBlezniuk, line 19: Let's create variable for duration

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+    padding: 23px 0;
+  }
+
+  &__item:first-child {
+    margin-left: 0;
+  }
+
+  &__item:last-child {
+    margin-right: 0;
+  }
```
OleksiiBlezniuk, line 30: Use sass nesting for pseudoclasses

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,56 @@
...
+  color: $styleColorMoyo;
+}
+
+.is-active::after {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+
+  content: "";
+
```
OleksiiBlezniuk, line 49: Use sass nesting for pseudoelements

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,24 @@
...
+
+@for $index from 1 through $stars-count {
+  .stars--#{$index} > .stars__star:nth-child(-n + #{$index}) {
+    background-image: url(../images/star-active.svg);
+  }
+}
+
+.stars > .stars__star:last-of-type {
+  margin-right: 0;
+}
```
OleksiiBlezniuk, line 24: let's avoid specificity increasing

src/styles/reset.scss
```diff
@@ -0,0 +1,7 @@
+body,
+p,
+ul,
+h2 {
+  margin: 0;
+  padding: 0;
+}
```
OleksiiBlezniuk, line 7: All of this element's don't have any default padding

OleksiiBlezniuk (changes requested):

src/styles/reset.scss
anastasiia-nahaiska: I found the information that ul have default padding-left: 40px
It would be better to write it separately from other resets like now or together?
image

OleksiiBlezniuk: Yeah, I was wrong about ul, sorry. Better to write it separately

OleksiiBlezniuk (approved):
Good job!

PR #1908, opened 2022-08-20
Develop

NikolayGit1:
- [DEMO LINK](https://nikolaygit1.github.io/layout_catalog/)
- [TEST REPORT LINK](https://nikolaygit1.github.io/layout_catalog/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,16 @@
+// #region card box
```
alexandra-protyanova, line 1: u don't need regions when u separate blocks by files

alexandra-protyanova (changes requested):
U margins between cards 59px, but should be 50px, that is about why tests haven't passed

PR #1947, opened 2022-08-23
Develop

SerhiiRomaniukWD:
- [DEMO LINK](https://SerhiiRomaniukWD.github.io/layout_catalog/)
- [TEST REPORT LINK](https://SerhiiRomaniukWD.github.io/layout_catalog/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
+.card {
+  display: flex;
+  flex-direction: column;
+
+  box-sizing: border-box;
+
+  width: 200px;
```
DenysKleimenov, line 7: I think it will be more correct to use a variable for the width of the card.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
...
+  padding: 32px 0 16px;
+  margin-right: 48px;
+  margin-bottom: 46px;
+
+  border: 1px solid $color-close-to-white;
+  border-radius: 5px;
```
DenysKleimenov, line 13: You should create a variable for the border-radius because we use it more than one time.

src/styles/utils/__fonts.scss
```diff
@@ -0,0 +1,17 @@
...
+  font-family: "Roboto";
+  src: url(../fonts/Roboto/Roboto-Medium.ttf);
+  font-weight: 500;
+}
+
+@font-face {
+  font-family: "Roboto";
+  src: url(../fonts/Roboto/Roboto-Regular.ttf);
+  font-weight: 400;
+}
```
DenysKleimenov, line 17: You could just add these fonts in HTML instead of creating a new file and typing excessive code, but it's not a big problem.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
...
+  &__image {
+    @include size-y-x(134px, 160px);
+
+    padding: 0 19px;
+    margin-bottom: 40px;
+  }
+
+  &__info {
+    margin-bottom: 4px;
+    padding: 0 16px;
```
DenysKleimenov, line 32: I think it would be easier to remove these paddings and add horizontal padding to the **.card** and change your padding size of **card__image**

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  &__link-button:hover {
+    color: $color-close-to-azure;
+
+    transition: 0.3s;
+
+    background-color: $color-close-to-white;
+  }
+}
```
DenysKleimenov, line 105: You should use nesting, put your **:hover** inside **&__link-button**

src/styles/blocks/main.scss
```diff
@@ -0,0 +1,10 @@
+.main {
+  display: flex;
+  flex-wrap: wrap;
+  box-sizing: border-box;
+  justify-content: center;
+
+  margin: 0 auto;
+  padding-top: 50px;
+  width: 1000px;
```
DenysKleimenov, line 9: Why did you set the width 1000px? I think you're supposed to calculate width with formula (($card-width * 4) + (card-right-padding * 3))

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,53 @@
...
+    padding: 23px 0;
+
+    color: $color-black;
+    text-decoration: none;
+  }
+
+  &__link:hover {
+    color: $color-close-to-azure;
+    transition: 0.3s;
+  }
```
DenysKleimenov, line 33:
```suggestion
&__link {
    display: block;

    padding: 23px 0;

    color: $color-black;
    text-decoration: none;
    
    &:hover {
    color: $color-close-to-azure;
    transition: 0.3s;
    }
  }
```

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,23 @@
...
+    background-position: center;
+  }
+
+  &--0 > &__star,
+  &--1 :not(:first-child),
+  &--2 > &__star:nth-child(n+3),
+  &--3 > &__star:nth-child(n+4),
+  &--4 :last-child {
+    background-image: url(../images/star.svg);
+  }
```
DenysKleimenov, line 22: I think you could use an array **@for** instead of this

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,53 @@
...
+      margin-right: 0;
+    }
+  }
+
+  &__link {
+    display: block;
+
+    padding: 23px 0;
+
+    color: $color-black;
```
DenysKleimenov, line 26: Link color shouldn't be black, it should be **#000e36** as I remember.

denyskleimenov (changes requested):

src/index.html
```diff
@@ -7,6 +7,471 @@
...
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#" class="nav__link is-active">apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="#" class="nav__link">samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="#" class="nav__link">smartphones</a>
+          </li>
```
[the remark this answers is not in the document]
SerhiiRomaniukWD: agree

src/styles/blocks/card.scss
SerhiiRomaniukWD: agree

src/styles/blocks/card.scss
SerhiiRomaniukWD: agree

src/styles/utils/__fonts.scss
SerhiiRomaniukWD: i know, but i like when fonts are as separate file

src/styles/blocks/card.scss
SerhiiRomaniukWD: It's normal idea but not for this case, cause image has another padding, I have better idea

src/styles/blocks/card.scss
SerhiiRomaniukWD: I agree

src/styles/blocks/nav.scss
SerhiiRomaniukWD: agree

src/styles/blocks/stars.scss
SerhiiRomaniukWD: good idea

src/styles/blocks/nav.scss
SerhiiRomaniukWD: in this case should be black, for another cases i have variable $color-close-to-black with #000e35 value

src/styles/blocks/main.scss
SerhiiRomaniukWD: Yes, good idea, I will fix it

src/styles/blocks/main.scss
```diff
@@ -0,0 +1,10 @@
+.main {
+  display: flex;
+  flex-wrap: wrap;
+  box-sizing: border-box;
+  justify-content: center;
+
+  margin: 0 auto;
```
Anna-Tkachenko, line 7: you can't set margins to BEM block

Anna-Tkachenko (changes requested):
Check correct color for the text in navigation menu. If you set correct color - it should fix some test

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,96 @@
...
+  &:nth-child(4n) {
+    margin-right: 0;
+  }
+
+  &:nth-last-child(-n+4) {
+    margin-bottom: 0;
+  }
+
+  &__image {
+    @include size-y-x(134px, 160px);
```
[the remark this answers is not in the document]
SerhiiRomaniukWD: ^ y
 |
 |                 
 |                 x
 ----------->

MaksimusCRS (approved):
GJ

PR #1949, opened 2022-08-23
Solution

Andrew-Skoryk:
- [DEMO LINK](https://Andrew-Skoryk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Andrew-Skoryk.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/nav.scss
+++ b/src/styles/blocks/nav.scss
@@ -20,6 +20,7 @@
 
   &__link {
     display: flex;
+    position: relative;
     padding: 1.9em 0;
 
     color: $text-main-color;
@@ -28,22 +29,21 @@
     &:hover {
       color: $button-color;
     }
-  }
-}
 
-.is-active {
-  position: relative;
-  color: $button-color;
+    &--is-active {
+      color: $button-color;
 
-  &::after {
-    position: absolute;
-    display: block;
-    bottom: 0;
-    height: 4px;
-    width: 100%;
+      &::after {
+        position: absolute;
+        display: block;
+        bottom: 0;
+        height: 4px;
+        width: 100%;
 
-    content: "";
-    background-color: $button-color;
-    border-radius: 8px;
+        content: "";
+        background-color: $button-color;
+        border-radius: 8px;
+      }
+    }
   }
 }
--- a/src/index.html
+++ b/src/index.html
@@ -18,7 +18,7 @@
       <nav class="nav header__nav">
         <ul class="nav__list">
           <li class="nav__item">
-            <a href="#apple" class="nav__link is-active">
+            <a href="#apple" class="nav__link nav__link--is-active is-active">
               Apple
             </a>
           </li>


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,7 @@
+.page {
+  font-family: "Roboto", Arial, sans-serif;
+  font-weight: 400;
```
dmytristepanov, line 3: Better to set the font family to the HTML tag, if you don't want to use a tag to set it, you can add class to the HTML tag

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,49 @@
...
+    line-height: 14px;
+
+    &:last-child {
+      margin: 0;
+    }
+  }
+
+  &__link {
+    display: flex;
+    padding: 1.9em 0;
```
dmytristepanov, line 23: Where did you get this magic number?

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,49 @@
...
+    display: block;
+    bottom: 0;
+    height: 4px;
+    width: 100%;
+
+    content: "";
+    background-color: $button-color;
+    border-radius: 8px;
+  }
+}
```
dmytristepanov, line 49: will be better to create active modifier for the links, and move these styles to those modifiers
and use is-active class only in html for tests passing

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
...
+  &__star {
+    margin-right: 4px;
+    @include starSize($star-size);
+
+    background-image: url(../images/star.svg);
+    background-position: center;
+    background-repeat: no-repeat;
+  }
+
+  @for $i from 1 through 5 {
```
dmytristepanov, line 14: I think to name variables more clearly is better for understanding what's going on here

dmytristepanov (commented):
Great work! Left few comments

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,16 @@
+.catalog {
+  @extend %space-between;
+  flex-wrap: wrap;
+  max-width: mainWidth($card-width, $space-between-cards);
+  margin: 0 auto;
```
Anna-Tkachenko, line 5: you can't set margins to BEM block

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,11 @@
+.header {
+  @extend %space-between;
+  padding: 0 50px;
+  margin: 0 auto;
```
Anna-Tkachenko, line 4: same here

src/styles/blocks/nav.scss
Anna-Tkachenko: 👀

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,49 @@
...
+  &__link {
+    display: flex;
+    position: relative;
+    padding: 1.9em 0;
+
+    color: $text-main-color;
+    @extend %link-transformation;
+
+    &:hover {
+      color: $button-color;
```
Anna-Tkachenko, line 30: let's add transition for this property

Anna-Tkachenko (approved):

PR #1951, opened 2022-08-23
Catalog was created

DenysKleimenov:
- [DEMO LINK](https://DenysKleimenov.github.io/layout_catalog/)
- [TEST REPORT LINK](https://DenysKleimenov.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,93 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+  border: 1px solid $white-border-color;
+  border-radius: $card-border-radius;
+  width: 200px;
+
+  &__info-logo {
+    width: 160px;
+    height: 134px;
+    padding: 0 3px;
+    margin-bottom: 40px;
+  }
+
+  &__info-title {
+    color: $dark-color;
+    font-family: $font-family-inherit;
+    font-weight: 500;
+    font-size: 12px;
+
+    line-height: 18px;
+    margin-bottom: 4px;
+  }
+
+  &__info-code {
+    font-family: $font-family-inherit;
+    font-size: 10px;
+    line-height: 14px;
+    color: $grey-color;
+    margin-bottom: 16px;
+  }
+
+  &__review {
+    @extend %flex-between-center;
+
+    line-height: 14px;
+    margin-bottom: 24px;
+
+    &-mark {
+      font-family: $font-family-inherit;
+      font-size: 10px;
+      line-height: 14px;
+    }
+  }
+
+  &__price {
+    @extend %flex-between-center;
+    margin-bottom: 16px;
+
+    &-text {
+      font-family: $font-family-inherit;
+      font-size: 12px;
+      line-height: $card-price-line;
+      color: $grey-color;
+    }
+
+    &-amount {
+      font-family: $font-family-inherit;
+      font-weight: 700;
+      font-size: 16px;
+      line-height: $card-price-line;
+      color: $dark-color;
+    }
+  }
+
+  &__button {
+    height: 38px;
+
+    color: $white-color;
+    background: $sea-color;
+
+    @extend %flex-center-center;
+    text-decoration: none;
+
+    font-weight: 700;
+    font-size: 14px;
+    font-family: $font-family-inherit;
+    line-height: 16px;
+    text-transform: uppercase;
+
+    border: 1px solid $sea-color;
+    border-radius: $card-border-radius;
+
+    transition: color 0.5s ease-out, background-color 0.5s ease-out;
+
+    &:hover {
+      background: $white-color;
+      color: $sea-color;
+    }
+  }
+}
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,0 +1,13 @@
+.header {
+  @extend %flex-between-center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__nav-bar {
+    margin-right: 50px;
+  }
+
+  &__logo {
+    margin: 10px 0 10px 50px;
+    @include headerLogoSize($header-logo-size);
+  }
+}
--- a/src/styles/blocks/nav.scss
+++ b/src/styles/blocks/nav.scss
@@ -1,0 +1,39 @@
+.nav {
+  &__list {
+    display: flex;
+    list-style: none;
+  }
+
+  &__link {
+    display: inline-block;
+    text-decoration: none;
+    color: #000e36;
+    text-transform: uppercase;
+    font-weight: 500;
+    font-size: 12px;
+    padding: 23px 0;
+    margin-left: 20px;
+    transition: color 0.4s;
+
+    &:hover {
+      color: $sea-color;
+    }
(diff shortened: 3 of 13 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
...
+  &__info-logo {
+    width: 160px;
+    height: 134px;
+    padding: 0 3px;
+    margin-bottom: 40px;
+  }
+
+  &__info-title {
+    color: $dark-color;
+    font-family: $font-family-inherit;
```
iPanda177, line 19: you already set a font-family in html tag? Is this really necessary to add a variable few times during your code?

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,39 @@
+.nav {
+  &__list {
+    display: flex;
+    list-style: none;
+  }
+
+  &__link {
+    display: inline-block;
+    text-decoration: none;
+    color: #000e36;
```
iPanda177, line 10: you can actually set this color by variable as well)

iPanda177 (approved):
GJ!!

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,87 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
```
OleksiiBlezniuk, line 4: Use * selector for box-sizing, usually you need to change logic of all elements sizing

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,87 @@
...
+    @extend %flex-between-center;
+
+    line-height: 14px;
+    margin-bottom: 24px;
+
+    &-mark {
+      font-size: 10px;
+      line-height: 14px;
+    }
+  }
```
OleksiiBlezniuk, line 43: `-mark` is not an element or modifier, so it's a wrong usage of nesting. Although it provides ability to store some class name as variable, we shouldn't use it for class name combining (it's not clear if it's nested class, or just other class with name that includes this word)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,87 @@
...
+    margin-bottom: 16px;
+
+    &-text {
+      font-size: 12px;
+      line-height: $card-price-line;
+      color: $grey-color;
+    }
+
+    &-amount {
+      font-weight: 700;
```
OleksiiBlezniuk, line 56: Looks like generic font-weight value, so let's create a variable

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,87 @@
...
+
+    font-weight: 700;
+    font-size: 14px;
+    line-height: 16px;
+    text-transform: uppercase;
+
+    border: 1px solid $sea-color;
+    border-radius: $card-border-radius;
+
+    transition: color 0.5s ease-out, background-color 0.5s ease-out;
```
OleksiiBlezniuk, line 80: Would be nice to create a variable for duration, then we can use it for every transition to make it's duration consistent with other page transitions

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,13 @@
+.header {
+  @extend %flex-between-center;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__nav-bar {
+    margin-right: 50px;
+  }
+
+  &__logo {
+    margin: 10px 0 10px 50px;
```
OleksiiBlezniuk, line 10: Strange solution
1 - you could make header flexible and align items in it (what if we decide to make links higher?)
2 - it's more logical to set paddings for parent container instead of margins for both of it's children

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,39 @@
...
+    display: flex;
+    list-style: none;
+  }
+
+  &__link {
+    display: inline-block;
+    text-decoration: none;
+    color: $link-color;
+    text-transform: uppercase;
+    font-weight: 500;
```
OleksiiBlezniuk, line 12: Looks like generic font-weight value, so let's create a variable

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,14 @@
...
+$grey-color: #616070;
+$white-border-color: #f3f3f3;
+$white-color: #fff;
+$card-border-radius: 5px;
+$products-width: 944px;
+$star-size: 16px;
+$card-price-line: 18px;
+$card-width: 200px;
+$product-card-margin-right: 48px;
+$font-family-inherit: inherit;
```
OleksiiBlezniuk, line 13: Do you really need this variable? Most of elements will inherit font-family by default

OleksiiBlezniuk (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,90 @@
+* {
+  box-sizing: border-box;
+}
```
OleksiiBlezniuk, line 3: It's strange that you set styles for every element in BEM-block styles file

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,90 @@
...
+      color: $grey-color;
+    }
+
+    &-amount {
+      font-weight: $bold-weight;
+      font-size: 16px;
+      line-height: $card-price-line;
+      color: $dark-color;
+    }
+  }
```
OleksiiBlezniuk, line 64: I've written about it during last review

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  &__price {
+    @extend %flex-between-center;
+    margin-bottom: 16px;
+
+    &-text {
+      font-size: 12px;
+      line-height: $card-price-line;
+      color: $grey-color;
+    }
```
OleksiiBlezniuk, line 56:
```suggestion
  &__price {
    @extend %flex-between-center;
    margin-bottom: 16px;
  }

  &__price-text {
    font-size: 12px;
    line-height: $card-price-line;
    color: $grey-color;
  }
```

OleksiiBlezniuk (changes requested):
GJ, but don't use sass nesting for combining words into class name

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,86 @@
...
+  padding: 32px 16px 16px;
+  border: 1px solid $white-border-color;
+  border-radius: $card-border-radius;
+  width: $card-width;
+
+  &__info-logo {
+    width: 160px;
+    height: 134px;
+    margin: 0 auto;
+    margin-bottom: 40px;
```
OleksiiBlezniuk, line 13: You can combine this rule

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,39 @@
...
+  &__link {
+    display: inline-block;
+    text-decoration: none;
+    color: $link-color;
+    text-transform: uppercase;
+    font-weight: $regular-weight;
+    font-size: 12px;
+    padding: 23px 0;
+    margin-left: 20px;
+    transition: color 0.4s;
```
OleksiiBlezniuk, line 16: Better to use same duration for all transitions on the page (you also created variable for this purpose)

OleksiiBlezniuk (approved):
GJ!

PR #2047, opened 2022-09-11
v01 Solution

home2ego911:
- [DEMO LINK](https://home2ego911.github.io/layout_catalog/)
- [TEST REPORT LINK](https://home2ego911.github.io/layout_catalog/report/html_report/)


marina-tilniak (changes requested):
if I delete one card it should works correct
image

PR #2063, opened 2022-09-18
Catalog was created

dudar-y:
- [DEMO LINK](https://dudar-y.github.io/layout_catalog/)
- [TEST REPORT LINK](https://dudar-y.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,103 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
+  display: flex;
+  flex-direction: column;
+  border: 1px solid $border-color;
+  border-radius: 5px;
+
+  &__image-block {
+    display: flex;
+    margin: 32px 19px 40px;
+  }
+
+  &__image {
+    width: 160px;
+    height: 134px;
+  }
+
+  &__text-block {
+    margin: 0 $card-text-margin 4px;
+  }
+
+  &__text {
+    font-weight: $medium-font;
+    font-size: 12px;
+    line-height: 18px;
+    color: $text-color-dark;
+  }
+
+  &__code-block {
+    margin: 0 $card-text-margin 16px;
+  }
+
+  &__code {
+    font-weight: $regular-font;
+    font-size: 10px;
+    line-height: 14px;
+    color: $text-color-light;
+  }
+
+  &__reviews-block {
+    display: flex;
+    justify-content: space-between ;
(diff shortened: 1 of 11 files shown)


alexandra-protyanova:
Снимок экрана 2022-09-19 в 14 41 42
If I add a card, nothing should be broken

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,103 @@
+.card {
+  box-sizing: border-box;
```
alexandra-protyanova, line 2: move to `*` selector

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,103 @@
...
+  width: 200px;
```
alexandra-protyanova, line 3: do not set external geometry to bem blocks, use mix instead

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,103 @@
...
+    justify-content: space-between ;
+    align-items: center;
+    margin: 0 $card-text-margin 24px;
+  }
+
+  &__reviews {
+    font-weight: $regular-font;
+    font-size: 10px;
+    line-height: 14px;
+    color: $text-color-dark;
```
alexandra-protyanova, line 52: u can move `html` selector for this property, and use `inherit` where needed

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,35 @@
+.products {
+  width: 100%;
```
alexandra-protyanova, line 2: here as well

alexandra-protyanova (changes requested):

src/styles/blocks/card.scss
dudar-y: We have very different font sizes and font styles for the elements on this page. If I don't change, will it be an error?

alexandra-protyanova:
Снимок экрана 2022-09-20 в 08 39 47
карточки должны добавляться с лева на право

src/styles/blocks/card.scss
alexandra-protyanova: Do we have one which me use more often, then it make sense to move to `html`, if not u can leave it

alexandra-protyanova (changes requested):
Need to fix adding of cards

alexandra-protyanova (approved):

PR #2107, opened 2022-10-04
Catalog

sofiia-dovhan:
- [DEMO LINK](https://sofiia-dovhan.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sofiia-dovhan.github.io/layout_catalog/report/html_report/)

tests fails on 1024px, i don't understant why

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,445 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link href="https://fonts.googleapis.com/css2?family=Inter&family=Prata&family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="https://www.moyo.ua/ua/">
+          <img src="./images/logo.png" alt="logo">
+        </a>
+      </div>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#apple">APPLE</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#samsung">SAMSUNG</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#smartphones">SMARTPHONES</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              data-qa="nav-hover"
+              href="#laptops"
+            >
+              LAPTOPS & COMPUTERS
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#gadgets">GADGETS</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#tablets">TABLETS</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#photo">PHOTO</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#video">VIDEO</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="main__container">
+        <div class="product-card" data-qa="card">
+          <div class="product-card__image">
+            <img
+              class="product-card__image-size"
+              src="./images/imac.jpeg"
+              alt="iMac"
+            >
+          </div>
+
+          <h1 class="product-card__title reset">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h1>
+          <p class="product-card__code reset">
+            Product code: 195434
+          </p>
+
+          <div class="stars-box product-card__stars-box">
+            <div class="stars-box__stars">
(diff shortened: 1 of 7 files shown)


src/styles/main-container.scss
```diff
@@ -0,0 +1,14 @@
...
+  justify-content: center;
+  box-sizing: border-box;
+
+  &__container {
+    display: flex;
+    justify-content: space-between;
+    flex-wrap: wrap;
+    box-sizing: border-box;
+    width: 950px;
+    margin: 50px 40px;
```
IOneStarI, line 12: in mockup we can see that 1024 and 1200 have different space (margins/paddings) between border of the page and a card 
i thing it should change

IOneStarI (commented):
GJ! You did nice work! 
just couple fixes.
I hope it will help you)

also  in Requirements cards should have fixed width and fixed distances between them
in your situation distance between cards depends on window size.

src/index.html
```diff
@@ -5,8 +5,445 @@
...
+
+          <div class="product-card__price">
+            <p class="product-card__price-title reset">Price:</p>
+            <p class="product-card__price-amount reset">$2,199</p>
+          </div>
+
+          <a
+            class="product-card__button"
+            href="#"
+            data-qa="card-hover"
```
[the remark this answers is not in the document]
sofiia-dovhan: Thank you, really very useful remark, i just missed that! :grin:

src/styles/main-container.scss
sofiia-dovhan: Thanks for review, but i think that  spaces between border and products should be changeble because of fixed spaces between cards and fixed cards container width 
:question:

IOneStarI: point from check list
13 [LAYOUT] - don't hardcode max-width, your catalog container should work fine for any number of cards
(here the is container width should not be fixed. )

there what i talking about. Different distance 
https://prnt.sc/vl02QLHDVACC for 1024px
https://prnt.sc/dv1abmdmGQAI for 1200px

sofiia-dovhan: i think i fixed this))

IOneStarI: Nice GJ! 
But it still breaks when screen is small. But don't worry you will fix this in future HW.
i think for this task it is ok))) lets see what is mentors will say

sofiia-dovhan: it doesn't break, i just want it to transfer on the next row if the screen is smaller than the width of 4 cards plus paddings
Soooooo let see)))
I think only mentor can answer)))
But thank you for your attention !)

src/styles/main.scss
```diff
@@ -1,3 +1,14 @@
...
+@import  "./navigation.scss";
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
+}
+
+.reset {
+  margin: 0;
 }
```
mykola-koval, line 14: better use a reset.css file instead of adding an extra class to all elements

src/styles/navigation.scss
```diff
@@ -0,0 +1,49 @@
...
+    list-style: none;
+  }
+
+  &__item {
+    display: flex;
+    padding-left: 20px;
+    font-weight: 500;
+  }
+
+  &__list > &__item:first-child {
```
mykola-koval, line 17: checklist point 3

src/styles/navigation.scss
```diff
@@ -0,0 +1,49 @@
...
+  &__link {
+    display: inline-block;
+    font-size: 12px;
+    text-decoration: none;
+    color: #060b35;
+    line-height: 60px;
+  }
+
+  &__link:hover {
+    color: #00acdc;
```
mykola-koval, line 30: this color is a good candidate for a variable

src/styles/product-card.scss
```diff
@@ -0,0 +1,94 @@
+.product-card {
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
```
mykola-koval, line 4: it is a good practice to add `box-sizing: border-box;` on all elements using `*`

src/styles/product-card.scss
```diff
@@ -0,0 +1,94 @@
...
+  font-size: 12px;
+
+  &__image {
+    display: flex;
+    justify-content: center;
+    text-align: center;
+    margin-top: 16px;
+  }
+
+  &__image-size {
```
mykola-koval, line 21: why use a separate class for image size? add this to main class

src/styles/product-card.scss
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  &__button:hover {
+    background-color: #fff;
+    color: #00acdc;
+    border-color: #00acdc;
+  }
+
+  &__stars-box {
+    margin-top: 16px;
```
mykola-koval, line 84: you can add bottom margin on prev element and remove this class

src/styles/stars.scss
```diff
@@ -0,0 +1,27 @@
...
+  &__star {
+    height: 16px;
+    width: 16px;
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: 50%;
+    margin: 0 4px 0 0;
+  }
+
+  &__stars > &__star:nth-child(-n+4) {
```
mykola-koval, line 24: make it work with any amount of stars, like in `stars` task

PR #2141, opened 2022-10-05
create a catalog

sssolomia:
- [DEMO LINK](https://sssolomia.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sssolomia.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,7 +1,3 @@
-.container__main {
-  margin: 0 auto;
-}
-
 .main {
   display: flex;
   flex-wrap: wrap;
--- a/src/styles/blocks/product-card.scss
+++ b/src/styles/blocks/product-card.scss
@@ -2,7 +2,7 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  width: 200px;
+  width: $product-card-width;
   padding: 32px 16px 16px;
   border: 1px solid #f3f3f3;
   border-radius: 5px;
--- a/src/styles/blocks/nav.scss
+++ b/src/styles/blocks/nav.scss
@@ -26,7 +26,7 @@
     font-size: 12px;
     line-height: 14px;
     text-transform: uppercase;
-    color: #000;
+    color: #060b35;
 
     &:hover {
       color: $color-blue;
--- a/src/styles/utils/__variables.scss
+++ b/src/styles/utils/__variables.scss
@@ -4,4 +4,6 @@ $color-white: #fff;
 
 $starSize: 16px;
 
+$product-card-width: 200px;
+
 $width-main: calc((200px * 4) + (48px * 3));


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,29 @@
+.container__main {
+  margin: 0 auto;
```
karyna-salnykova, line 2: i think  there's no need to use 'auto'

src/styles/blocks/reset.scss
```diff
@@ -0,0 +1,15 @@
...
+  font-family: "Roboto", sans-serif;
+  color: #060b35;
+}
+
+body,
+h2,
+p {
+  margin: 0;
+  padding: 0;
+}
```
karyna-salnykova, line 11: in my opinion, it`d be better to use classes instead of tags, and add one class for all elements you need to reset

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,13 @@
...
+  align-items: center;
+  height: 60px;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__logo {
+    height: 40px;
+    margin-right: 50px;
+  }
+}
```
andrii-bassarab, line 13: Nav links color is not black any more. That's why the header tests don't pass.

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,98 @@
+.product-card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: 200px;
```
andrii-bassarab, line 5: it is better to set the width value by a variable, because it can change

src/styles/blocks/catalog.scss
sssolomia: you're right! it isn't necessary here

src/styles/blocks/header.scss
sssolomia: thank you!you helped me a lot

src/index.html
```diff
@@ -4,9 +4,421 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link 
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet"
```
mykola-koval, line 10:
```suggestion
      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" 
      rel="stylesheet"
```

src/index.html
```diff
@@ -4,9 +4,421 @@
...
+          </li>
+
+          <li class="nav__item">
+            <a 
+              href="#laptops&computers" 
+              class="nav__link"
+              data-qa="nav-hover"
+            >
+            Laptops & Computers
+            </a>
```
mykola-koval, line 42:
```suggestion
            >
              Laptops & Computers
            </a>
```

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,25 @@
+.main {
+  display: flex;
+  flex-wrap: wrap;
+  max-width: $width-main;
+  padding: 50px 40px;
+  box-sizing: initial;
```
mykola-koval, line 6: why u need this?

sssolomia: because without this my cards won't be positioned correctly. but I will try to change it because it isn't quite right.

PR #2167, opened 2022-10-15
Catalog

Illia0308:
- [DEMO LINK](https://Illia0308.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Illia0308.github.io/layout_catalog/report/html_report/)


SanyaBratashchuk (changes requested):
It shouldn't be broken if I add more cards 
image

PR #2229, opened 2022-11-02
catalog

anastasia-radchencko:
https://anastasia-radchencko.github.io/layout_catalog/
https://anastasia-radchencko.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,112 @@
+.cards {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  box-sizing: border-box;
+  justify-content: center;
+  padding-top: 50px;
+
+  &__body {
+    width: 200px;
+    margin-left: 24px;
+
+    &:not(:nth-child(4n)) {
+      margin-right: 24px;
+    }
+
+    &:nth-last-child(-n + 4) {
+      margin-bottom: 46px;
+    }
+  }
+}
+
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+
+  background-color: $standart-color;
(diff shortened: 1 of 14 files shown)


DarkMistyRoom (changes requested):
Good, but pay attention to this point 
image
image

and it should look like this when you add a new item
image

SanyaBratashchuk (changes requested):
image

src/styles/normalize.scss
```diff
@@ -0,0 +1,8 @@
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
```
stanislav-topikha, line 6: add fallback font

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,115 @@
...
+    width: $card-width;
+
+    &:nth-child(4n) {
+      margin-right: 0;
+    }
+
+    &:nth-child(-n + 4) {
+      margin-top: 0;
+    }
+  }
```
stanislav-topikha, line 114: No external positioning for BEM blocks, event through modifiers, move to element and mix to card

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,115 @@
...
+  flex-direction: column;
+  justify-content: center;
+
+  background-color: $standart-color;
+  border: 1px solid $card-border-color;
+  border-radius: $main-border-radius;
+
+  padding: 0 16px 16px;
+
+  &-container {
```
stanislav-topikha, line 12: what is this? BEM? 
image
why does it located outside of block?

stanislav-topikha (changes requested):

PR #2244, opened 2022-11-05
Catalog

DmytroRozenberh:
- [DEMO LINK](https://DmytroRozenberh.github.io/layout_catalog/)
- [TEST REPORT LINK](https://DmytroRozenberh.github.io/layout_catalog/report/html_report/)


SanyaBratashchuk (changes requested):
image
image

ivanchuk-yurii (changes requested):
comments from [product-card](https://github.com/mate-academy/layout_product-cards/pull/2069) are also applicable here. commit all your `scss` files so we can review them

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
+.card {
+  box-sizing: border-box;
```
TarasHoliuk, line 2:
```suggestion
```

it will be inherited from `*` selector

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,58 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+  
+.header__logo {
+  width: 40px;
+  height: 40px;
+  margin: 10px 50px;
+}
+  
+.nav__list {
```
TarasHoliuk, line 15: since nav is BEM-block it should be in a separate file

TarasHoliuk (approved):
GJ! I've added a few comments (just FYI)

PR #2261, opened 2022-11-07
Develop

Lun4trick:
- [DEMO LINK](https://lun4trick.github.io/layout_catalog/)
- [TEST REPORT LINK](https://lun4trick.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,10 +3,442 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="unknow">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a href="index.html">
+        <img
+          src="images/logo.png"
+          alt="logo"
+          class="logo"
+      >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="apple" class="nav__link is-active">apple</a>
+          </li>
+          <li class="nav__item">
+            <a href="samsung"class="nav__link">samsung</a>
+          </li>
+          <li class="nav__item">
+            <a href="smartphones"class="nav__link">smartphones</a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="laptops & computers"
+              class="nav__link"
+              data-qa="nav-hover">laptops & computers
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="gadgets"class="nav__link">gadgets</a>
+          </li>
+          <li class="nav__item">
+            <a href="tablets"class="nav__link">tablets</a>
+          </li>
+          <li class="nav__item">
+            <a href="photo"class="nav__link">photo</a>
+          </li>
+          <li class="nav__item">
+            <a href="video"class="nav__link">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="products">
+      <div class="card" data-qa="card">
+        <img
+          src="images/imac.jpeg"
+          alt="product-img"
+          class="card__product"
+        >
+
+        <div class="card__specs">
+          <div class="card__name">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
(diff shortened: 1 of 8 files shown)


src/index.html
```diff
@@ -3,10 +3,442 @@
...
+          src="images/logo.png"
+          alt="logo"
+          class="logo"
+      >
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="apple" class="nav__link is-active">apple</a>
+          </li>
```
lefteris-L, line 28: you should space these so it's clear where they begin and end

src/index.html
```diff
@@ -3,10 +3,442 @@
...
+              </div>
+            </div>
+
+            <div class="card__pricing-price">
+              $2,199
+            </div>
+          </div>
+          <div class="card__link" data-qa="card-hover">
+            <label>
+              <a href="Buy" class="card__link">BUY</a>
```
lefteris-L, line 101: no need to wrap with label

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,127 @@
+.card {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
```
lefteris-L, line 4: you can replace all this with padding

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,127 @@
...
+
+    background-color: $linkColor;
+    border-radius: 5px;
+    color: #fff;
+    border: 1px solid $linkColor;
+    /* transition: background-color 1s, color 1s; */
+  }
+
+  &__link:hover {
+    background-color: #fff;
```
lefteris-L, line 110: you can just put this in the &__link { } and write it as &:hover

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,127 @@
...
+    background-repeat: no-repeat;
+  }
+
+  &__star:nth-child(n+2) {
+    margin-left: 4px;
+  }
+
+  @for $rating from 1 through 5 {
+    &--#{$rating} .stars__star:nth-child(-n + #{$rating}) {
+      background-image: url(/src/images/star-active.svg);
```
lefteris-L: Very good

lefteris-L (changes requested):
You should use paddings on the card container instead of space-around to have more control over your elements, and margin on the elements to control the distance between them. flex direction column is the default without flex so in most cases you don't need it. If you see a selector twice like &__link and &__link:hover, you should nest it. . Your checks should pass if you fix your margins/paddings and remove the extra wrappers that may be causing spacing problems. You really should space multi-line siblings in html to make your code easier to navigate. Sorry for the long review. Solid work all around.

src/styles/blocks/card.scss
lefteris-L: not the `box-sizing: border-box;`

Lun4trick:
> You should use paddings on the card container instead of space-around to have more control over your elements, and margin on the elements to control the distance between them. flex direction column is the default without flex so in most cases you don't need it. If you see a selector twice like &__link and &__link:hover, you should nest it. . Your checks should pass if you fix your margins/paddings and remove the extra wrappers that may be causing spacing problems. You really should space multi-line siblings in html to make your code easier to navigate. Sorry for the long review. Solid work all around.

I think im done with the changes what you requested, lmk if i missed something, all of them all relevant ideas so thank you! :)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,135 @@
+.card {
```
meforoma, line 1: i'm bit concerned with overall card construct
would it be much easier if your entire card has `padding: 32px 16px 16px`
so all the children are already limited and it simplifies a lot of styling

lefteris-L (approved):
Just a note, block classes, like .nav and .logo should be in their own scss file. Great job!

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,135 @@
...
+
+  &__name {
+    color: $mainColor;
+    font-size: 12px;
+    font-weight: $fontWeight-medium;
+    line-height: 150%;
+  }
+
+  &__code {
+    margin: 4px 0 17px;
```
meforoma, line 35: let's remind that we should apply either margin-top or -bottom across our entire styles
now some elements have no margin at all, while others have both top and bottom - fix this too

meforoma (changes requested):
@Lun4trick let's fix few comments, probably more will follow

src/index.html
```diff
@@ -69,382 +69,379 @@
           alt="product-img"
```
meforoma, line 69: think of disabled people who can't see images
and your shop has more than one product, if each image regardless of the product will have `alt="product-img"` - not great
why not `alt="APPLE iMac 27 inches Retina 5K Monoblock"`?

PR #2277, opened 2022-11-08
inital commit

44shaggy:
- [DEMO LINK](https://44shaggy.github.io/layout_catalog/)
- [TEST REPORT LINK](https://44shaggy.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,406 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <nav class="nav-list">
+        <li class="nav-list__item">
+          <a class="nav-list__link nav-list__link--active" href="#Apple">Apple</a>
+        </li>
+
+          <li class="list-item">
+            <a class="link" href="#Samsung">Samsung</a>
+          </li>
+
+        <li class="nav-list__item">
+          <a class="nav-list__link" href="#Smartphones">Smartphones</a>
+        </li>
+
+          <li class="list-item">
+            <a class="link" data-qa="hover" href="#Laptops&Computers">Laptops & Computers</a>
+          </li>
+
+          <li class="list-item">
+            <a class="link" href="#Gadgets">Gadgets</a>
+          </li>
+
+          <li class="list-item">
+            <a class="link" href="#Tablets">Tablets</a>
+          </li>
+
+          <li class="list-item">
+            <a class="link" href="#Photo">Photo</a>
+          </li>
+
+          <li class="list-item">
+            <a class="link" href="#Video">Video</a>
+          </li>
+        </menu>
+      </nav>
+      <a class="logo" href="#Home">
+        <img src="./images/logo.png" alt="logo">
+      </a>
+    </header>
+    <main class="page">
+      <div class="card">
+        <img class="card__image" src="./images/imac.jpeg" alt="apple computer">
+
+        <h1 class="card__product-name">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
+
+        <p class="card__product-code">
+          Product code: 195434
+        </p>
+
+        <div class="card__reviews-group">
+          <div class="stars stars--4">
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+            <span class="stars__star"></span>
+          </div>
+
+          <span class="card_reviews">
+            Reviews: 5
+          </span>
+        </div>
+
+        <div class="card__price-group">
+          <span class="card__price">
+            Price:
+          </span>
+
+          <span class="card__tag">
+            $2,199
+          </span>
+        </div>
+
+        <a
+          class="card__button"
+          href="#"
+          data-qa="hover"
+        >
+          <div class="card__button-name">
+            Buy
+          </div>
+        </a>
+      </div>
+      <div class="card">
+        <img class="card__image" src="./images/imac.jpeg" alt="apple computer">
+
+        <h1 class="card__product-name">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
+
+        <p class="card__product-code">
+          Product code: 195434
+        </p>
+
+        <div class="card__reviews-group">
+          <div class="stars stars--4">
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -4,9 +4,424 @@
...
   </head>
-  <body>
-    <h1>Catalog</h1>
+  <body class="page">
+    <header class="header">
+      <nav class="nav-list">
+        <li class="nav-list__item">
+          <a class="nav-list__link nav-list__link--active is-active" href="#Apple">Apple</a>
+        </li>
+
```
amitpatiljc, line 15: understand the list stucture
```
<ul> -- list
  <li> -- list item (1)
    <a> -- link
  </li>
  <li> -- list item (1)
    <a> -- link
  </li>
  ...
</ul>
```

amitpatiljc (changes requested):
Structure HTML first suggested in the comment

src/index.html
44shaggy: Got it

src/styles/blocks/logo.scss
```diff
@@ -0,0 +1,3 @@
+.logo {
+  height: 40px;
```
amitpatiljc, line 2: instead of hard coded height, you can use display flex and align it to centre vertically

amitpatiljc (changes requested):

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,7 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  box-sizing: border-box;
+  padding: 0 50px;
+}
```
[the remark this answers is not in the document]
44shaggy: > check in the mock design there is mock design for the header 
I noticed I missed the box-shadow property. Is this what you meant?

> also consider min and max width here as this is parent container, you can control child elements from here
Could you elaborate? What benefits would using those properties have?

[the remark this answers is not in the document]
amitpatiljc: 1. yes I meant box-shadow property, sorry
2. with  min and max width you can achieve responsiveness, you can avoid hard coded values

src/index.html
```diff
@@ -1,12 +1,430 @@
...
+    </header>
+    <main class="main-content page__main">
+      <div class="card" data-qa="card">
+        <img
+          class="card__image"
+          src="./images/imac.jpeg"
+          alt="apple computer"
+        >
+
```
amitpatiljc, line 62: I think you should add one more container like this for all the cards
```
<main. class="main-content page__main">
   <div class="cards"> (add this container as a parent to all 8 cards)
      <div class="card" data-qa="card">
```

and consider this checklist point , 
```
there should always be 4 cards in a row (not 2, 3 or 5)
use :nth-child(4n) to select every 4th element
use :nth-last-child(-n + 4) to select 4 last elements
```

amitpatiljc (changes requested):

src/styles/blocks/header.scss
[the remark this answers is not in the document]
44shaggy: > 2. with  min and max width you can achieve responsiveness, you can avoid hard coded values

Did I hard code values in the header? Seems to me it takes 100% container width while nav width is dictated by width of its grandchild elements.

src/index.html
44shaggy: > I think you should add one more container like this for all the cards

Done. It doesn't change the test results though. What purpose does it serve?

> use :nth-child(4n) to select every 4th element
> use :nth-last-child(-n + 4) to select 4 last elements

Not sure what I'm trying to achive here? Constant number of .card elements in a row is already defined by container width. Also I used :nth-child(4n) selector.

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    &:hover {
+      color: $color-blue-accent;
+    }
+
+    &--active {
+      color: $color-blue-accent;
+
+      &::after {
+        position: absolute;
```
amitpatiljc, line 31: this is absolute to which element or class?
i think it should be to `nav__link`

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,40 @@
...
+        align-self: flex-end;
```
amitpatiljc, line 32: it is `position: absolute` you need to position it with bottom, top or left ...

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,40 @@
...
+        background-color: $color-blue-accent;
+        height: 4px;
+        content: "Apple";
```
amitpatiljc, line 35: ``
```suggestion
        content: "";
```

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,40 @@
...
+        color: rgba(0, 0, 0, 0);
```
amitpatiljc, line 36: are you sure this is the right color

amitpatiljc (changes requested):

44shaggy:
Hey @amitpatiljc . The changes you requested refer to code I reused from the Header task. That task was approved.  I'm confused. Should I change reused code which was already approved?

jstmpelowycz:
> * DEMO LINK

sure you can

src/styles/blocks/nav.scss
44shaggy: It's resposible for a blue bar under Apple nav-link--active. My understanding was it should only be under Apple link. No?

44shaggy: Got it.

src/styles/blocks/nav.scss
44shaggy: Got it.

src/styles/blocks/main-content.scss
```diff
@@ -0,0 +1,6 @@
+.main-content {
+  // display: flex;
+  // flex-wrap: wrap;
+  width: $main-width;
+  padding: 1px;
```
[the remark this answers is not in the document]
44shaggy: I agree, it looks silly. I was trying to apply this task requirement:
"- cards container should have fixed paddings (see 1024px example), bottom the same as top"

Which container does it refer to?

polosanya (changes requested):
Try to set 48px gap between cards. On design it's 50px, but without borders, so probably that's the reason why tests fail
catalog

qoa11a (approved):

PR #2280, opened 2022-11-08
passed test

Alireza-Mohammadhossein:


.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
```
UmizDemud, line 2: I am not used to seeing linthtmlrc.json here...

Before you push with "git push origin develop"
You add using git add [something here]

In these tasks and in general you only push the code you write or added. So

git add src
- would only add the code you wrote in the src folder

git add readme.md
- would only add the changes in the readme file

it seems like you do

git add .
- which adds everything, kinda inefficient this way.

Let me know if this helps

UmizDemud: so 

git add src && git add readme.md

should be the right way

UmizDemud (approved):
Looks great, I'll pass it for the mentor review.

Altough, please check before you push next time about what to add in the commit. You can text for more info.

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,136 @@
...
+    line-height: 18px;
+    text-align: right;
+    color: $accent-color;
+  }
+
+  /* #endregion */
+
+  /* #region Product button */
+
+  &__btn {
```
meforoma, line 106: instead of regions you could just name your classes more explicit
```suggestion

  &__button {
```

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,136 @@
...
+
+  &__price-value {
+    font-size: 16px;
+    font-weight: 700;
+    line-height: 18px;
+    text-align: right;
+    color: $accent-color;
+  }
+
+  /* #endregion */
```
meforoma, line 102: regions again are pretty much redundant
given that scss enables you to nest styles:
so you will have an explicit collapsible region without the need to mark regions at all
```suggestion

  &__price {
    display: flex;
    justify-content: space-between;
    align-items: center;
    margin-bottom: 16px;
    
    &-text {
      font-size: 12px;
      font-weight: 400;
      line-height: 18px;
      text-align: left;
      color: $secondary-color;
    }
    
    &-value {
      font-size: 16px;
      font-weight: 700;
      line-height: 18px;
      text-align: right;
      color: $accent-color;
    }
  }
```

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,136 @@
...
+
+  &__review-rate {
+    font-size: 10px;
+    font-weight: 400;
+    line-height: 14px;
+    color: $accent-color;
+    text-align: right;
+  }
```
meforoma, line 73: nest it inside `&__review {`
like this
```suggestion
  &__review {
    &-rate {
      font-size: 10px;
      font-weight: 400;
      line-height: 14px;
      color: $accent-color;
      text-align: right;
    }
  }

```

meforoma (changes requested):
good job 👍 
let's make it better with proper styles nesting - see comments and apply throughout the code.

And:
Monosnap Catalog 2022-11-10 15-14-11

src/index.html
```diff
@@ -4,9 +4,421 @@
...
+          <li class="nav__item">
+            <a href="#" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+              data-qa="nav-hover"
+              >
```
naziamoff, line 39: closing brace should be at the same level as the opening one

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,134 @@
...
+      text-align: right;
+      color: $accent-color;
+    }
+  }
+
+  /* #endregion */
+
+  /* #region Product button */
+
+  &__button {
```
naziamoff, line 104: to me it seems that the button should be a separate block, because it is logically different from the card layout itself.

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,134 @@
...
+      font-weight: 400;
+      line-height: 14px;
+      color: $accent-color;
+      text-align: right;
+    }
+  }
+
+  /* #region Product starts */
+
+  .stars {
```
naziamoff, line 53: same with stars

naziamoff (approved):
LGTM 🔥 
But you should separate logically different pieces of of layout into blocks in order to make the code reusable and structure more clear

PR #2293, opened 2022-11-09
add catalog

NikitaNevv:
https://NikitaNevv.github.io/layout_catalog/
https://NikitaNevv.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -380,6 +380,7 @@
             buy
           </a>  
         </div>
+        
       </div>
     </main>
   </body>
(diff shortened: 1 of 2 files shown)


DarkMistyRoom (changes requested):
- It shouldn't go over the screen

image

- And it should work properly on adding a card
image

SanyaBratashchuk (changes requested):
It shouldn't be broken if I add more cards 
image

Bohdan-Maksymiuk (approved):

PR #2301, opened 2022-11-09
Catalog task done!

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/layout_catalog/)
- [TEST REPORT LINK](https://mmahfuzi.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,413 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <!-- linthtml-configure attr-req-value="false" -->
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+    href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+    rel="stylesheet"
+    crossorigin
+    >
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+
+      <a class= "container" href="#">
+        <img
+        src="https://raw.githubusercontent.com/mmahfuzi/layout_moyo-header-en/master/src/images/logo.png"
+        alt="logo">
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#samsung">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#smartphones">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              data-qa="nav-hover"
+              href="#laptops & computers"
+            >
+              Laptops & Computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#gadgets">Gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#tablets">Tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#photo">Photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#video">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <div class="cards-container">
+      <div class="cards-container__page page">
+        <div class="card-block">
+          <main data-qa="card" class="card">
+            <img
+              class="card__image image"
+              src="./images/imac.jpeg"
+              alt="Laptop"
+            >
+
+            <div class="card__text">
+              <p class="card__text card__text--1">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</p>
+              <p class="card__text card__text--2">Product code: 195434</p>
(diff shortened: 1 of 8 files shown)


CharAnagn (changes requested):
Hey, there is something wrong with your Test Report link, check for any typo and I will review it again, to see where I can help you. Also try to add the readme.md to your commit :)

src/index.html
```diff
@@ -5,8 +5,413 @@
...
+            <a class="nav__link" href="#video">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <div class="cards-container">
+      <div class="cards-container__page page">
+        <div class="card-block">
+          <main data-qa="card" class="card">
```
CharAnagn, line 72: After the header tag you should have one main tag that wraps the main content of the page, for this example the whole catalog(cards), and wrap each card with the div. Even better if you use section tag after the main and wrap each card with article tag for semantic purposes.

src/index.html
```diff
@@ -5,8 +5,413 @@
...
+              <div class="card__rate-review">Reviews: 5</div>
+            </div>
+
+            <div class="card__price">
+              <div class="card__price-text">
+                Price:
+              </div>
+
+              <div class="card__price-num">$2,199</div>
+            </div>
```
CharAnagn, line 102:
```suggestion
            <p class="card__price">
              <span class="card__price-text">
                Price:
              </span>
              <span class="card__price-num">
                 $2,199
               </span>
            </p>
```
Not a mistake, but since it's basically a text split in two, you can wrap it in a p tag and use span for each text. This way you can tell what's the content of this block and for semantic purposes.

src/styles/blocks/cards-container.scss
```diff
@@ -0,0 +1,18 @@
+$paddingX: 40px;
+
+@function getWidth($card-width) {
+  @return 4 * $card-width + 2 *$paddingX + 3 * 48;
+}
```
CharAnagn, line 5: It's better to create a new file _functions.scss and put your functions there, so it can be easier for you to find them

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,132 @@
...
+$color-blue: #00acdc;
+$card-width: 200px;
+
+.card {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  align-items: flex-start;
+  width: #{$card-width};
```
CharAnagn, line 12:
```suggestion
  width: $card-width;
```
Is this syntax you have something else than a variable? Cause for a variable it's less code to use $variable-name.

I believe this syntax is to reference the variable you declared in a for loop. But for the variables you declared in your variables.scss you can use this syntax.

CharAnagn (changes requested):
Other than the comment in the html, it looks super good! :)

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,7 @@
+.page {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
meforoma, line 5: it does the job for the current task,
yet @mmahfuzi when need to stick with a design measures - it's better to provide the 
explicit `_px` spacing

mmahfuzi: > it does the job for the current task, yet @mmahfuzi when need to stick with a design measures - it's better to provide the explicit `_px` spacing

Thanks for your comment.
Sure, if I was allowed to use 'gap' I would add explicit px and was easier to handle.

meforoma: @mmahfuzi i meant that you could use `margin-right: 48px` ))

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,126 @@
...
+
+  background: white;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  box-sizing: border-box;
+
+  &__image {
+    display: block;
+    height: 134px;
+    margin: 16px auto 0;
```
meforoma, line 18: Monosnap Catalog 2022-11-10 16-34-40

src/index.html
```diff
@@ -5,8 +5,415 @@
...
+              <img
+                class="card__image image"
+                src="./images/imac.jpeg"
+                alt="Laptop"
+              >
+
+              <div class="card__text">
+                <p class="card__text card__text--1">APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)</p>
+                <p class="card__text card__text--2">Product code: 195434</p>
+              </div>
```
meforoma, line 83: Monosnap Catalog 2022-11-10 16-38-11

src/index.html
```diff
@@ -5,8 +5,415 @@
...
+                  <div class="stars__star"></div>
+                  <div class="stars__star"></div>
+                  <div class="stars__star"></div>
+                  <div class="stars__star"></div>
+                </div>
+
+                <div class="card__rate-review">Reviews: 5</div>
+              </div>
+
+              <p class="card__price">
```
meforoma, line 97: Monosnap Catalog 2022-11-10 16-40-06

meforoma (changes requested):
please apply changes discussed

src/index.html
meforoma: this one is still there @mmahfuzi 
this is the last one 🙏

meforoma (changes requested):
you haven't made one agreed change

mmahfuzi:
> you haven't made one agreed change

Ohhh sorry I missed that

meforoma (approved):

PR #2305, opened 2022-11-09
if someone finds that one error I buy him/her a coffee

CharAnagn:
- [DEMO LINK](https://CharAnagn.github.io/layout_catalog/)
- [TEST REPORT LINK](https://CharAnagn.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/cards.scss
+++ b/src/styles/blocks/cards.scss
@@ -1,8 +1,7 @@
 .cards {
   display: flex;
   flex-wrap: wrap;
-
   width: 1024px;
-
+  justify-content: center;
   padding: 50px 40px;
 }
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,5 +1,5 @@
 .catalog {
   display: flex;
-  width: 1024px;
-  margin: 0 auto;
+  min-width: 1024px;
+  justify-content: center;
 }


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,106 @@
+.card {
+  box-sizing: border-box;
+  width: $cardWidth;
+  padding: 32px 16px 16px;
+
+  margin-right: 48px;
+  margin-bottom: 46px;
+
```
mmahfuzi, line 8: margin is an external style and should be inside elements not blocks. (to follow BEM)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,106 @@
...
+    background-color: $colorMain;
+    border: 1px solid $colorMain;
+    border-radius: 5px;
+
+    transition: color 0.3s, background-color 0.3s;
+
+    &:hover {
+      color: $colorMain;
+      background-color: #fff;
```
mmahfuzi, line 103: I see you have used color #fff three times in the card block, isn't it better to use it as a variable as you did for $colorMain?

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,9 @@
+.cards {
+  display: flex;
+  flex-wrap: wrap;
+
+  width: 1024px;
+
+  margin: 0 auto;
```
mmahfuzi, line 7: Again margin in a block (.cards). Should be moved to an element. I see there are other the same kinda mistake in other blocks as well.

mmahfuzi (commented):
Good job! Except for the margins in the blocks the rest of the code is nice and correct.

src/styles/blocks/card.scss
CharAnagn: Yeah unfortunately this margin is because we cannot use gap for the flexbox.

src/styles/blocks/cards.scss
CharAnagn: I removed this margin, didn't need it apparently to center :D

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,5 @@
+.catalog {
+  display: flex;
+  width: 1024px;
+  margin: 0 auto;
+}
```
meforoma, line 5: try adjusting style a bit to pass the last test - your entire component should be 1200px wide on respective screen
```
.catalog {
    display: flex;
    min-width: 1024px;
    justify-content: center;
}
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,106 @@
...
+    text-align: center;
+    text-transform: uppercase;
+    color: #fff;
+    text-decoration: none;
+
+    background-color: $colorMain;
+    border: 1px solid $colorMain;
+    border-radius: 5px;
+
+    transition: color 0.3s, background-color 0.3s;
```
meforoma, line 99: very nice to transition certain props instead of all
👍

meforoma (changes requested):
love it 👍 
let's try to pass the final test and you're all good

naziamoff (approved):
🔥

PR #2311, opened 2022-11-09
Develop

Kavster92:
- [DEMO LINK](https://Kavster92.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Kavster92.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/cards.scss
+++ b/src/styles/blocks/cards.scss
@@ -1,0 +1,117 @@
+.cards-container {
+
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+  width: 1024px;
+  max-width: 1200px;
+  margin: 0 auto;
+  padding: 50px 40px;
+}
+
+.card {
+
+  width: 200px;
+  margin-right: 48px;
+  margin-bottom: 46px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  border: 1px solid var(--card-border);
+  border-radius: 5px;
+  color: #060b35;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 18px;
+
+  &:nth-child(4n) {
+    margin-right: 0;
+  }
+
+  &:nth-last-child(-n + 4) {
+    margin-bottom: 0;
+  }
+
+  &__image-wrapper {
+    margin-bottom: 35px;
+    text-align: center;
+  }
+
+  &__image {
+    width: 160px;
+    height: 134px;
+  }
+
+  &__title {
+    margin-bottom: 4px;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+  }
+
+  &__product-code {
+    margin: 0 0 16px;
+    color: var(--card-fallback);
+    font-size: 10px;
+    font-weight: 400;
+    line-height: 14px;
+  }
+
+  &__reviews {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    margin-bottom: 24px;
+    font-size: 10px;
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,27 @@
+@import "./blocks/header.scss";
+@import "./blocks/cards.scss";
+@import "./blocks/stars.scss";
+
+:root {
+  --card-main: #060b35;
+  --card-fallback: #616070;
+  --card-secondary: #00acdc;
+  --card-border: #f3f3f3;
+}
```
CharAnagn, line 10: You should use sass. Create a new folder called utils and inside it a new file called variables and declare your variables there.

$variable-name: value;

src/styles/main.scss
```diff
@@ -1,3 +1,27 @@
...
   margin: 0;
 }
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
```
CharAnagn, line 27: This one aswell in a seperate file inside blocks.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,48 @@
+.nav {
+
+  &__list {
+    margin: 0;
+    padding: 0;
+    display: flex;
+    list-style: none;
+  }
+
```
CharAnagn, line 9: It's better to put the whole nav block and all the children element in a seperate file like navigation.scss inside blocks for better structure.

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,117 @@
...
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  border: 1px solid var(--card-border);
+  border-radius: 5px;
+  color: #060b35;
+  font-family: "Roboto", sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 18px;
+
```
CharAnagn, line 28: Same here, this folder is called cards.scss, and its the container of the single card. So you can put it in a separate file.

CharAnagn (changes requested):
I think you forgot to commit your index.html. You have the same test as me that didn't pass and I could figure out why :D Just a bit rearrangement with the files and folder structure and you are perfect!!

naziamoff (changes requested):
Before sending the task on mentor review pls make sure that you have fixed comments from the student's review pls 🙏 

If you are not sure whether the comments are valid - you can specify the point from the student that reviewed them or ask mentors for it if needed

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,117 @@
+.cards-container {
+
```
naziamoff, line 2: you don't need an empty space between selector and its rules

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,117 @@
...
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+  width: 1024px;
+  max-width: 1200px;
+  margin: 0 auto;
+  padding: 50px 40px;
+}
+
+.card {
```
naziamoff, line 14: the card looks like a separate block to me. Why don't we make it a separate file as well?
This approach is usually used for making styling easier, it's gonna be easier to navigate between the pieces of your code if you need to change something

src/styles/blocks/header.scss
naziamoff: yeah, I agree with @CharAnagn about this point

naziamoff (approved):
Well done! 
But would be even better if you separated the blocks to the separate files. It really increases the maintainability of your code.
What's more - except for colors you can add variables for the font size, it's a really common practice too

PR #2313, opened 2022-11-09
Layout catalog was added at the first time

hyrych-max:
- [DEMO LINK] - https://hyrych-max.github.io/layout_catalog/
- [TEST REPORT LINK] - https://hyrych-max.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -82,7 +82,9 @@
     border: 1px solid transparent;
     border-radius: 5px;
 
-    transition: 0.3s;
+    transition: color $button-card-transition,
+      border $button-card-transition,
+      background-color $button-card-transition;
 
     &:hover {
       color: $blue-color;
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -7,6 +7,7 @@ $border-color: #f3f3f3;
 $card-width: 200px;
 $card-margin: 48px;
 $cards-in-row: 4;
+$button-card-transition: 0.5s;
 
 $min-padding: 40px;
 $max-padding: 128px;


src/index.html
```diff
@@ -5,8 +5,371 @@
...
+            BUY
+          </a>
+        </div>
+      </article>
+
+      <article class="card page__item" >
+        <img
+          class="card__product-img"
+          src="images/imac.jpeg"
+          alt="product_photo">
```
loralevitska, line 144:
```suggestion
          alt="product_photo"
        >
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,96 @@
+.card {
+  width: $card-width;
+  padding: 32px 16px 16px;
+
+  font-weight: 400;
+
+  border: 1px solid #f3f3f3;
```
loralevitska, line 7: it is a good practice to make all colors variable

src/styles/blocks/navigation.scss
```diff
@@ -0,0 +1,59 @@
+.nav {
+  display: block;
+
+  &__list {
+    display: flex;
+    justify-content: space-between;
```
loralevitska, line 6: these properties are often repeated and can be transferred to mixin

loralevitska (changes requested):
1. All advice from the [fe-chat](https://mate-parttime.slack.com/archives/C041JA5EJKV/p1668031747431449) is valid, do not forget to correct them 
2. Pay attention for my comments

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+    font-weight: 700;
+    text-decoration: none;
+    text-align: center;
+    color: $card-bgc;
+
+    background-color: $blue-color;
+    border: 1px solid transparent;
+    border-radius: 5px;
+
+    transition: 0.3s;
```
polosanya, line 85: It is better to clarify for which properties you want to apply the transition.
https://developer.mozilla.org/en-US/docs/Web/CSS/transition
Also, transition time should be the same for the whole page, so it's better to add a variable for it.

polosanya (changes requested):
Almost perfect! Please check my comment

PR #2317, opened 2022-11-09
completed

jackelashkar:
- [DEMO LINK](https://jackelashkar.github.io/layout_catalog/)
- [TEST REPORT LINK](https://jackelashkar.github.io/layout_catalog/report/html_report/)


src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,21 @@
...
+  &--1 &__star:nth-child(-n+1),
+  &--2 &__star:nth-child(-n+2),
+  &--3 &__star:nth-child(-n+3),
+  &--4 &__star:nth-child(-n+4),
+  &--5 &__star:nth-child(-n+5) {
+    background-image: url(../images/star-active.svg);
+  }
+  & &__star:nth-child(5) {
+    background-image: url(../images/star.svg);
+  }
```
meforoma, line 20: seem almost identical with Ala implementation - so same comment - use a loop according to a checklist

meforoma (changes requested):
same comments as for Ala

and make the cards container fixed at 1024px, so that on narrower screens cards don't wrap & clash and there's a horizontal padding
narrower screens

alaaashkar (approved):
good job ! looks good to me !

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  @for $starCount from 1 to 5 {
+    &--#{$starCount} &__star:nth-child(-n+#{$starCount}) {
+      background-image: url(../images/star-active.svg);
+    }
+  }
+
+  & &__star:nth-child(5) {
+    background-image: url(../images/star.svg);
+  }
```
meforoma, line 20: what does this part do?
what if i want `class="stars stars--5"` - i expect it to show 5 active stars.
It does not show me 5 stars (((

meforoma (changes requested):
the active-stars comment - i've recognised it recently - let's fix it :)

src/styles/blocks/stars.scss
jackelashkar: i fixed it.

jackelashkar:
> 

fixed also

src/styles/blocks/stars.scss
jackelashkar: fixed

meforoma (approved):
🎉

PR #2332, opened 2022-11-10
Catalog

IoghiStudio:
- [DEMO LINK](https://IoghiStudio.github.io/layout_catalog/)
- [TEST REPORT LINK](https://IoghiStudio.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -1,12 +1,366 @@
...
+              Video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main>
+      <h1 class="visually-hidden">Catalog</h1>
+
```
Alireza-Mohammadhossein, line 91: I think it is better to remove the tag that you want to hide it.
Visibility hidden is better for dynamic content that you want to show them after some actions like hover...

src/index.html
```diff
@@ -1,12 +1,366 @@
...
+          </p>
+
+          <a
+            data-qa="card-hover"
+            class="card__button"
+            href="#"
+          >
+            Buy
+          </a>
+        </article>
```
Alireza-Mohammadhossein, line 130: I am not sure, but I think section and article are not good choices for these cards especially for SEO, just div is better choice

src/styles/utils/visually-hidden.scss
```diff
@@ -0,0 +1,11 @@
...
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  margin: -1px !important;
+  border: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip: rect(0,0,0,0) !important;
+  white-space: nowrap !important;
+}
```
Alireza-Mohammadhossein, line 11: why did you use these styles? just try don't use !important

Alireza-Mohammadhossein (commented):
Everything was perfect, especially using sass :)
well done bro

src/styles/utils/visually-hidden.scss
IoghiStudio: Just for that class , so screen readers can still see the h1 , like misha did.

src/index.html
IoghiStudio: For semantics .

meforoma (approved):
nicey-nice 👍

PR #2336, opened 2022-11-11
Catalog has been created

NgineZZ:
- [DEMO LINK](https://NgineZZ.github.io/layout_catalog/)
- [TEST REPORT LINK](https://NgineZZ.github.io/layout_catalog/report/html_report/)


src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,103 @@
...
+      margin-right: 0;
+    }
+
+    &:nth-last-child(-n + 4) {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.card {
```
naziamoff, line 26: it's a separate block so it should better be in a separate file

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,103 @@
...
+    font-size: 12px;
+  }
+
+  &__price {
+    font-size: 16px;
+    font-weight: 700;
+    color: $main-text-color;
+  }
+
+  &__button {
```
naziamoff, line 81: Button could be a separate block as well I guess

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,103 @@
...
+    color: $main-text-color;
+  }
+
+  &__code {
+    margin-bottom: 16px;
+  }
+
+  &__review-container {
+    @include flex-aligning;
+    justify-content: space-between;
```
naziamoff, line 56: looks like when you use this mixin you always add the `justify-content: center;` rule. Do you think it would be reasonable to add it to the mixin as well?

naziamoff (approved):
Good job 🔥 
There are some comments to consider though

PR #2345, opened 2022-11-18
index.html and main.scss and style.css was added

VadymTsyndra:
https://VadymTsyndra.github.io/layout_catalog/
https://VadymTsyndra.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,385 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+
+    <header class="header">
+      <div class="header__logo">
+        <a href="#home">
+          <img src="./images/logo.png" alt="logo">
+        </a>
+      </div>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__element">
+            <a class="nav__link is-active" href="#home">Apple</a>
+          </li>
+
+          <li class="nav__element">
+            <a class="nav__link" href="#home">Samsung</a>
+          </li>
+
+          <li class="nav__element">
+            <a class="nav__link" href="#home">Smartphones</a>
+          </li>
+
+          <li class="nav__element">
+            <a
+              class="nav__link"
+              href="#home"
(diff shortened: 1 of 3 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,99 @@
...
+    background-color: $color-blue;
+    padding: 12px 70px 12px 69px;
+    text-transform: uppercase;
+    font-weight: 700;
+    font-size: 14px;
+    line-height: 16px;
+    text-align: center;
+    text-decoration: none;
+    color: $color-white;
+
```
Viktor-Kost, line 84: Add a `transition` for hover effect. It will be more user friendly 
([link](https://developer.mozilla.org/en-US/docs/Web/CSS/transition))

Sth like this `transition: color 250ms linear`

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  &__link {
+    text-decoration: none;
+    color: $color-dark;
+    padding-top: 22.5px;
+    padding-bottom: 22.5px;
+    position: relative;
+
```
Viktor-Kost, line 27: The same with transition

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,99 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  background-color: $color-white;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  width: 200px;
+  height: 408px;
+  margin-right: 48px;
```
Viktor-Kost, line 10: I see you apply margins for giving a space between cards. 
It's a bit hard to maintain. 
I recommend you use such rule as "gap" for parent element with a class `.main__card`

Read, it's easy) [link](https://developer.mozilla.org/en-US/docs/Web/CSS/gap)

src/index.html
```diff
@@ -5,8 +5,385 @@
...
+          <li class="nav__element">
+            <a class="nav__link" href="#home">Video</a>
+          </li>
+
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="main__card">
```
Viktor-Kost, line 64: Better use another class name for this element. Will we have secondary_card? You can call, for example, main_cards or something. 
It should be specific

Viktor-Kost (changes requested):
Nice!)
Apply some reommendations

SanyaBratashchuk (approved):

PR #2354, opened 2022-11-21
initial commit

ihorivna007:
- [DEMO LINK](https://ihorivna007.github.io/layout_catalog/)
- [TEST REPORT LINK](https://ihorivna007.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -7,6 +7,13 @@
   border-radius: 5px;
   padding: 32px 16px 16px;
 
+  &--position {
+    margin-bottom: 46px;
+    &:not(:nth-child(4n)) {
+      margin-right: 48px;
+    }
+  }
+
   &__description {
     margin-top: 40px;
   }
@@ -14,11 +21,11 @@
   &__image {
     display: flex;
     justify-content: center;
+  }
 
-    .image__imac {
-      height: 134px;
-      width: 160px;
-    }
+  &__logo {
+    height: 134px;
+    width: 160px;
   }
   
   &__name {
--- a/src/styles/blocks/nav.scss
+++ b/src/styles/blocks/nav.scss
@@ -12,28 +12,33 @@
   
   &__list-a {
     height: 60px;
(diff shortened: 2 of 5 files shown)


SanyaBratashchuk (changes requested):
image

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,99 @@
...
+
+  &__description {
+    margin-top: 40px;
+  }
+  
+  &__image {
+    display: flex;
+    justify-content: center;
+
+    .image__imac {
```
DarkMistyRoom, line 18: an element can't contain another element, according to BEM. Why don't set height and width to &__image?

src/styles/reset.scss
```diff
@@ -0,0 +1,6 @@
+body,
+h2,
+p {
+  margin: 0;
+  font-family: "Roboto", Helvetica, sans-serif;
```
DarkMistyRoom, line 5: Better set font-family to `html `selector and inherit where needed

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,39 @@
...
+    text-transform: uppercase;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+
+    &::after,
+    &::before {
+      content: "";
+      height: 4px;
+    }
```
DarkMistyRoom, line 26: Why do you need it? Set all styles to .is-active::after

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,39 @@
...
+    
+    &:hover,
+    &.is-active {
```
DarkMistyRoom, line 29: is active should be an independent class. Just remove `&`

DarkMistyRoom (changes requested):
The catalog should work properly on adding more cards. Better give up using row blocks, just common container for all cards
image

qoa11a (approved):

PR #2358, opened 2022-11-21
Develop

YevheniiHura:
- [DEMO LINK](https://YevheniiHura.github.io/layout_catalog/)
- [TEST REPORT LINK](https://YevheniiHura.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,233 @@
-body {
+$color-main-title: #060b35;
+$color-main-subtitle: #616070;
+$color-blue-theme: #00acdc;
+
+$card-right-margin: 48px;
+$card-bottom-margin: 46px;
+$card-width: 166px;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,233 @@
-body {
+$color-main-title: #060b35;
```
liliia-livandovskaya, line 1: checklist p.9 [BEM] - create a separate file per each BEM block styles that have the same name as the block

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments and create a separate file per each BEM block style

src/styles/main.scss
YevheniiHura: done)

liliia-livandovskaya (approved):
Great job! 🔥

PR #2440, opened 2022-12-10
Catalog created

VKonoplianyi:
Tried many ways to connect the stars svg, but nothing with success...
image

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,97 @@
+.card {
+  width: $card-width;
+  display: flex;
+  flex-direction: column;
+  padding: 32px 16px 16px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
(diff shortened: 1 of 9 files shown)


nazarmatsevych (changes requested):
Try to figure out with pathes or ask for help in fe_chat

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
...
+      color: $secondary-text-color;
+    }
+
+    &-value {
+      font-size: $font-size-l;
+      font-weight: $font-weight-l;
+      line-height: 18px;
+      color: $main-text-color;
+    }
+  }
```
DarkMistyRoom, line 72: Don't do it, because it can make a mess. better make separate selectors without nesting for each element
`&__price-section` and so on

SanyaBratashchuk (approved):

PR #2445, opened 2022-12-11
completed product catalog task added markup, and styles ( SCSS ) according to BEM,

GarryRocksman:
- [DEMO LINK](https://GarryRocksman.github.io/layout_catalog/)
- [TEST REPORT LINK](https://GarryRocksman.github.io/layout_catalog/report/html_report/)

have a nice day

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,70 +1,70 @@
-  .header {
-    max-width: 1200px;
-    left: 0;
-    top: 0;
-    margin: 0 auto;
-    padding: 0 50px;
+.header {
+  max-width: 1200px;
+  left: 0;
+  top: 0;
+  margin: 0 auto;
+  padding: 0 50px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background: rgb(255, 255, 255);
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__logo {
+    width: 40px;
+    height: 40px;
+  }
+
+}
+
+.nav {
+  &__list {
+    padding: 0;
     display: flex;
+    flex-direction: row;
     justify-content: space-between;
-    align-items: center;
-    background: rgb(255, 255, 255);
-    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  }
 
-    &__logo {
-      width: 40px;
-      height: 40px;
+  &__item {
+    display: flex;
+    padding: 0 10px;
+
+    &:first-child {
+      padding-left: 0;
+    }
+
+    &:last-child {
+      padding-right: 0;
     }
 
   }
 
-  .nav {
-    &__list {
-      padding: 0;
-      display: flex;
-      flex-direction: row;
-      justify-content: space-between;
-    }
+  &__link {
+    padding: 23px 0;
+    position: relative;
+    display: inline-block;
+    text-decoration: none;
+    text-transform: uppercase;
+    cursor: pointer;
+    color: $main-font-color;
 
-    &__item {
-      display: flex;
-      padding: 0 10px;
-
-      &:first-child {
-        padding-left: 0;
-      }
-
-      &:last-child {
-        padding-right: 0;
-      }
-    }
-
-    &__link {
-      padding: 23px 0;
-      position: relative;
-      display: inline-block;
-      text-decoration: none;
-      text-transform: uppercase;
-      cursor: pointer;
-      color: $main-font-color;
-
-      &:hover {
-        color: $blue-accent-color;
-      }
-    }
-
-  }
-
-  .is-active {
-    color: rgb(0, 172, 220);
-
-    &::before {
-      content: "";
-      height: 4px;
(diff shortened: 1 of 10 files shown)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
+.header {
+  max-width: 1200px;
+  left: 0;
+  top: 0;
```
Sergyurch, line 4: Why do you use these properties here?

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__logo {
+    width: 40px;
+    height: 40px;
+  }
+
+}
+
+.nav {
```
Sergyurch, line 20: Each block should be in a separate file.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
...
+    display: flex;
+    padding: 0 10px;
+
+    &:first-child {
+      padding-left: 0;
+    }
+
+    &:last-child {
+      padding-right: 0;
+    }
```
Sergyurch, line 38: Why do you use flex here? It does nothing in this case.
It's logically better to use margin-right here except for the last item.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  }
+
+  &__link {
+    padding: 23px 0;
+    position: relative;
+    display: inline-block;
```
Sergyurch, line 45: This property isn't needed in this case.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,70 @@
...
+    text-decoration: none;
+    text-transform: uppercase;
+    cursor: pointer;
```
Sergyurch, line 48: Cursor is pointer by default for links. You don't have to set it here.

src/styles/main.scss
```diff
@@ -1,3 +1,23 @@
...
+  font-style: normal;
+  font-weight: 500;
+  font-size: $fontSize;
+  color: $main-font-color;
+
+}
+
+.main {
+  display: flex;
+  flex-direction: column;
```
Sergyurch, line 21: Why do you use column here and not row?

src/styles/main.scss
```diff
@@ -1,3 +1,23 @@
...
+
+html {
   margin: 0;
+  font-family: Roboto, Arial, Helvetica, sans-serif;
+  font-style: normal;
+  font-weight: 500;
+  font-size: $fontSize;
+  color: $main-font-color;
+
+}
```
Sergyurch, line 17: Create 'page' block for html tag and move font properties there.

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,12 @@
...
+
+  @if $cardsNumber/4 < 1 {
+    $max-width: $cardsNumber
+      * ($cardWidth)
+      + ($cardsMargin * ($cardsNumber - 1))
+      + (2 * $catalogPadding);
+  }
+
+  @return $max-width;
+}
```
Sergyurch, line 12: Can't understand what you calculate here. We have a requirement in the task 'there should always be 4 cards in a row (not 2, 3 or 5)'. The number of cards in the row changes in you case if we change the width of viewport.
Just give your catalog fixed width (4 cards + margins between them).

Sergyurch (changes requested):
Try to improve the code. I left some comments.

src/styles/blocks/header.scss
GarryRocksman: I thought that header should be  BEM block

src/styles/blocks/header.scss
GarryRocksman: I have used flex for style decoration reset for the list, I agree it's hardcode so I will replace it with list-style: none ) Paddings, just forgot to replace it with margin.  Thanks.

src/styles/utils/_functions.scss
GarryRocksman: // here calculated max-width if the number of cards is less than 4, 
13 [LAYOUT] - Use sass functions for dynamic max-width calculation, your catalog container should work fine for any number of cards.

alexandra-protyanova:
Снимок экрана 2022-12-12 в 19 55 15
new card should be added from left to right with correct margins

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,88 @@
+.product-card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  box-sizing: border-box;
```
alexandra-protyanova, line 5: move to `*` selector to prevent code duplicating

alexandra-protyanova (approved):
Good job!

GarryRocksman:
> Снимок экрана 2022-12-12 в 19 55 15
> 
> new card should be added from left to right with correct margins



> 

fixed it

PR #2447, opened 2022-12-11
problem with px

Kocha1991:
- [DEMO LINK](https://Kocha1991.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Kocha1991.github.io/layout_catalog/report/html_report/)


DarkMistyRoom (changes requested):
- Check the data-qa attributes, they have new names
- Remove left indent on the nav
image

- Check the font styles. if you have question, ask in the chat
image

- fix the top padding in the main
image


- and gaps between cards

src/main.scss
```diff
@@ -0,0 +1,19 @@
...
+@import "./styles/logo.scss";
+@import "./styles/nav.scss";
+@import "./styles/is-active.scss";
+@import "./styles/card.scss";
+@import "./styles/starts.scss";
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
DarkMistyRoom, line 14: Don't use * selector for zeroing out your margins or paddings. It's still inefficient for browser to read your web document

DarkMistyRoom (approved):
Be consistent with your margins. Add only left or right, don't add both.

PR #2501, opened 2022-12-20
first commit

DmytroHorshunov:
https://DmytroHorshunov.github.io/layout_catalog/
https://DmytroHorshunov.github.io/layout_catalog/report/html_report/


SanyaBratashchuk (changes requested):
image

qoa11a (approved):

PR #2505, opened 2022-12-22
catalog develop

Anatolii-Polishchuk:
https://Anatolii-Polishchuk.github.io/layout_catalog/
https://Anatolii-Polishchuk.github.io/layout_catalog/report/html_report/


qoa11a (changes requested):
image
Layout shouldn't be broken if I add more cards.

image
Also add data-qa attributes.

SanyaBratashchuk (changes requested):
still broken if add more cards 
image

Virass (approved):

PR #2516, opened 2022-12-25
SASS / Catalog

valntyn:
**[DEMO LINK](https://valntyn.github.io/layout_catalog/)
[TEST REPORT LINK](https://valntyn.github.io/layout_catalog/report/html_report/)**

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,83 @@
+.card {
+  background-color: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: $card-paddings;
+  width: $card-width;
+
+  &__img {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin-bottom: 40px;
+    margin-left: 3px;
+  }
+
+  &__name {
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $color-general-text;
+    margin-bottom: 4px;
+  }
+
+  &__code-item {
+    color: $color-secondary-text;
+    font-size: 10px;
+    line-height: 14px;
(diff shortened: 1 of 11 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,83 @@
...
+    font-weight: 700;
+    font-size: 16px;
+    line-height: 18px;
+    color: $color-general-text;
+  }
+
+  &__button-buy {
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
witflash, line 66: Do you really need to use the link as `flex`? I think you may add just `text-align: center`.

witflash (changes requested):
Hey! You have to do layout without wrapping:
> use :nth-child(4n) to select every 4th element
> use :nth-last-child(-n + 4) to select 4 last elements 

But your layout looks strange on different screens:
image

src/styles/blocks/card.scss
valntyn: thank you for the correction!
I've changed parent's width from `max-width` to `width`
right now always 4 cards in row. 
also, I've centered  text 'buy' using `display: block` and `text-align: center`

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,82 @@
...
+    background-color: $color-general;
+    border: 1px solid $color-general;
+    border-radius: 5px;
+    text-transform: uppercase;
+    font-weight: 700;
+    font-size: 14px;
+    color: #fff;
+
+    &:hover {
+      background-color: #fff;
```
polosanya, line 78: Would be great to create variables for all the colors, so then you can change them in one place

polosanya (approved):
Good job!

PR #2567, opened 2023-01-12
products-catalog

mxwolfy:
- [DEMO LINK](https://mxwolfy.github.io/layout_catalog/)
- [TEST REPORT LINK](https://mxwolfy.github.io/layout_catalog/report/html_report/)


src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,17 @@
+.cards {
+  margin: 0 auto;
+  padding: 50px 0;
+  display: flex;
+  flex-wrap: wrap;
+  width: $cards-width;
+}
+
+.card--position {
```
witflash, line 9: It should be `.cards__card` element, not a `card` modifier. So you will have an element which may have margin properties.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
```
witflash, line 43: It's better to use a percentage or multiplier value for line height.
`line-height: 1.4`, `line-height: 150%`, etc.

witflash: And check all such places, please

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,88 @@
...
+    font-size: 16px;
+    line-height: 18px;
+    color: $main-color;
+  }
+
+  &__btn {
+    height: 40px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
witflash, line 72: There is no need to make the button as flexbox. You may do smth like this:
image

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,13 @@
+$main-color: #060b35;
+$second-color: #00acdc;
+$border-color: #f3f3f3;
+$third-color: #616070;
+$fourth-color: #fff;
```
witflash, line 5: It's better to place `color` as prefix:
```
$color-main
$color-border
etc.
```

witflash (changes requested):
Good work!
Try to improve your code, please

IvanFesenko (approved):
Well done

mxwolfy:
> Well done

Thanks

PR #2581, opened 2023-01-16
First commit

Khalniepiesova:
- [DEMO LINK](https://Khalniepiesova.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Khalniepiesova.github.io/layout_catalog/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,33 @@
...
+  padding-bottom: 71px;
+}
+
+.card:nth-child(-n+3) {
+  margin-right: 48px;
+}
+
+.row:nth-child(-n+1) {
+  margin-bottom: 46.5px;
 }
```
tetiana-ivanchenko, line 33: if this styles for blocks - it should have separate file, if it's element inside some block, it should have corresponding name

tetiana-ivanchenko (changes requested):

SanyaBratashchuk (changes requested):
image

SanyaBratashchuk (approved):

PR #2584, opened 2023-01-17
Undertake layout catalog task

Volodymyr-Terzov:
DEMO LINK - https://Volodymyr-Terzov.github.io/layout_catalog/
TEST REPORT LINK - https://Volodymyr-Terzov.github.io/layout_catalog/report/html_report/


src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,11 @@
...
+  margin: 0;
+}
+
+h2 {
+  margin: 0;
+}
+
+p {
+  margin: 0;
+}
```
DarkMistyRoom, line 11:
```suggestion
body, h2, p {
  margin: 0;
}
```

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,12 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background: $color-white;
+  box-shadow: 0 2px 4px $color-grey-nav-box-shadow;
+
+  .logo-image {
```
DarkMistyRoom, line 9: name it as element

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,17 @@
+@import "./utils/_functions.scss";
```
DarkMistyRoom, line 1: why do you need to import it here if you import it in the main file?

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,17 @@
...
+    height: 16px;
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: center;
+
+    &:not(:last-of-type) {
+      margin-right: 4px;
+    }
+  }
+}
```
DarkMistyRoom, line 17: reuse whole stars block. it should work with modifiers

DarkMistyRoom (changes requested):
everything should work properly on adding new card
image

src/styles/blocks/stars.scss
Volodymyr-Terzov: I just left it as a reminder (in case if the 'stars' block is further used as a separate block, it reminds that there is also a 'function' which applies to it).

src/index.html
```diff
@@ -1,12 +1,459 @@
...
+            <h2 class="card__product-name">
+              APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+            </h2>
+            <p class="card__product-code">
+              Product code: 195434
+            </p>
+
+            <div class="card__stars-container">
+              <div class="stars">
+                <div class="stars stars__stars--4">
```
loralevitska, line 73: Fix all cases

```suggestion
                <div class="stars stars--4">
```

src/styles/blocks/stars.scss
loralevitska: No need to save, something for the future. First, git has versions and you will not lose the code that is important to you, secondly, business tasks can change and unused code may never be used, but it can interfere and disorient.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,91 @@
...
+    font-size: 14px;
+    line-height: 16px;
+    color: $color-white;
+
+    &:hover {
+      background: $color-white;
+      border: 1px solid $color-blue-accent;
+      border-radius: 5px;
+      color: $color-blue-accent;
+      // padding: 11px 70px 11px 69px;
```
loralevitska, line 88: Remove comments

loralevitska (changes requested):
Keep up the good work! Some comments to fix:
Remove empty unused files or use it
image

SanyaBratashchuk (approved):

PR #2679, opened 2023-02-04
first attempt page was created

olyakachmarchyk:
https://olyakachmarchyk.github.io/layout_catalog/
https://olyakachmarchyk.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,12 +1,530 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#home">
+          <img src="images/logo.png" alt="MOYO logo">
+        </a>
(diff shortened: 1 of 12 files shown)


src/index.html
```diff
@@ -1,12 +1,529 @@
...
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+
+    <main>
+      <div class="catalog">
+        <section class="card catalog__card" data-qa="card">
+          <section class="picture card__picture">
```
ivanchuk-yurii, line 79: too many sections, here and lower `div` is enough)

src/index.html
```diff
@@ -1,12 +1,529 @@
...
+            <img
+              class="picture__img"
+              src="images/imac.jpeg"
+              alt="imac"
+            >
+          </section>
+
+          <section class="card__item item">
```
ivanchuk-yurii, line 87: redundant block, only `.stars` should be separate BEM block, everything else should be `.card`'s element

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,29 @@
...
+  display: flex;
+  align-items: center;
+
+  &__star {
+    width: 16px;
+    height: 16px;
+    background-image: url("/images/star-active.svg");
+    margin-right: 4px;
+    background-repeat: no-repeat;
+    background-position: center;
```
ivanchuk-yurii, line 23: it's better to combine into `background` property

ivanchuk-yurii (changes requested):
if I add a new card it should work properly 
image

VitaliyBondarenko1982 (approved):
Approve.
Just pay attention on scss syntax in next tasks

PR #2730, opened 2023-02-21
created catalog page

vpdrabynko:
https://vpdrabynko.github.io/layout_catalog/
https://vpdrabynko.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,340 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -5,8 +5,340 @@
...
-    <h1>Catalog</h1>
+    <header class="container">
+      <a href="#home" class="logo">
+        <img src="images/logo.png" alt="MOYO logo">
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#" class="nav__link is-active">
```
anstsot, line 21: it would be great to add different links as `#apple`, `#samsung` etc

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,12 @@
+.container {
+  display: flex;
+  justify-content: space-between;
+  margin: 0 auto;
+  align-items: center;
+}
```
anstsot, line 6: this style doesn't belong to header. it would be great to create separate scss file for it

anstsot (changes requested):
Looks good but there are just few last things to change

src/styles/blocks/header.scss
vpdrabynko: but i have header with class container
where should i move it then?
normalize.scss or split block into logo and container?

polosanya (approved):

PR #2814, opened 2023-03-01
catalog

myplekan:
https://myplekan.github.io/layout_catalog/
https://myplekan.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,7 +11,7 @@
   </head>
   <body>
     <header class="header">
-      <a href="#" class="logo">
+      <a href="#" class="header__logo">
         <img src="images/logo.png" alt="MOYO">
       </a>
       <nav class="nav">
@@ -20,12 +20,18 @@
             <a href="#apple" class="nav__link is-active">
               Apple
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#samsung" class="nav__link">
               Samsung
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#smartphones" class="nav__link">
               Smartphones
             </a>
+          </li>
+          <li class="nav__item">
             <a 
               href="#laptopsComputers" 
               class="nav__link" 
@@ -33,15 +39,23 @@
             >
               Laptops & Computers
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#gadgets" class="nav__link">
               Gadgets
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#tablets" class="nav__link">
               Tablets
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#photo" class="nav__link">
               Photo
             </a>
+          </li>
+          <li class="nav__item">
             <a href="#video" class="nav__link">
               Video
             </a>
@@ -70,7 +84,7 @@
         </h2>
         
         <div class="card__action">
-          <div class="stars stars--1">
+          <div class="stars stars--4">
             <div class="stars__star"></div>
             <div class="stars__star"></div>
             <div class="stars__star"></div>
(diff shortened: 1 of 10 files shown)


src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  &__buy:hover {
+    color: #00acdc;
+    background-color: #fff;
+    border: 1px solid #00acdc;
+  }
+}
+
+.stars {
```
vlad-demchuk, line 89: separate file

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,120 @@
...
+    background-image: url(../images/star-active.svg);
+    background-repeat: no-repeat;
+    background-position: center;
+  }
+
+  &__star:not(:last-child) {
+    margin-right: 4px;
+  }
+
+  &__star:nth-child(n+5) {
```
vlad-demchuk, line 103: stars should have modifiers as in the previous tasks
image

vlad-demchuk (changes requested):

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,77 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 15px;
+  margin-right: 48px;
```
slavanosatskyi, line 6: try to avoid external geometry for block elements, you can move to container__card element

src/index.html
```diff
@@ -4,9 +4,399 @@
...
+        <img src="images/logo.png" alt="MOYO">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
```
slavanosatskyi, line 24: add spaces between sibling elements
```suggestion
          </li>
          
          <li class="nav__item">
```

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,42 @@
...
+    text-decoration: none;
+  }
+
+  &__item:not(:last-child) {
+    margin-right: 17px;
+  }
+
+  &__link:hover {
+    color: #00acdc;
+  }
```
slavanosatskyi, line 25: you can use scss nesting here 
`&__link {
      ....
      &:hover {
         ....
      }
  }
`

slavanosatskyi (changes requested):
Nice work, just a couple of fixes needed. Also take a look on your grid when there are less than 8 elements 
Screenshot 2023-03-02 at 17 40 33

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,77 @@
...
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 15px;
+  border-radius: 5px;
+  background-color: #fff;
+  border: 1px solid #f3f3f3;
+  
+  &:nth-child(4n) {
+    margin-right: 0;
+  }
```
nazarmatsevych, line 12: Why do you need it?

I guess you can remove it and increase max-width for container or just use sass function and calculate max-width

nazarmatsevych (changes requested):

src/styles/blocks/cards.scss
myplekan: if I understand correctly, the test requires a max-width of 1024px and if there are less than 8 elements, then additional properties will need to be set for the second row

nazarmatsevych: yeah, but now it looks strange on screens smaller then 1024px
image

nazarmatsevych (approved):

PR #2833, opened 2023-03-05
add catalog

yura-enttsel:
- [DEMO LINK](https://yura-enttsel.github.io/layout_catalog/)
- [TEST REPORT LINK](https://yura-enttsel.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -1,0 +1,9 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+html {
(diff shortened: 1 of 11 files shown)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,9 @@
+* {
+  margin: 0;
```
vlad-demchuk, line 2: don't reset margin for all elements

vlad-demchuk (approved):
GJ!

PR #2866, opened 2023-03-13
Task done, but last failing test is invalid

artem-ivashchenko:
- [DEMO LINK](https://artem-ivashchenko.github.io/layout_catalog/)
- [TEST REPORT LINK](https://artem-ivashchenko.github.io/layout_catalog/report/html_report/)

The problem is that test doesn't approve my solution even if it's correct. I was using devtools and noticed that everything is ok, but test is failing. I applied to Q&A, couldn't solve the problem. I was told to send it anyway.
Please, check photos of the problem
https://drive.google.com/drive/folders/1E7yPjYT3H3JYE-Vu0Idf_SQBUB85lzCY?usp=sharing


src/styles/blocks/container.scss
```diff
@@ -0,0 +1,22 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  width: 1200px;
+  padding: 50px 128px;
```
agniya-i, line 5:
```suggestion
  width: <calculate value here>;
  padding-top: 50px;
  margin: auto; 
```

It can help with test: 
1. Do not hardcode width, it should base on the card width and gap between cards (should be about 944px)
2. Centering div using` margin:auto;`  (not it looks like not centered on page), and after that you don't need to set paddings

agniya-i (changes requested):

src/index.html
```diff
@@ -6,7 +6,327 @@
...
+          src="images/logo.png"
+          alt="Moyo__logo"
+          class="header__logo-pic"
+        >
+      </a>
+
+      <nav class="nav header__nav">
+        <ul class="nav__list">
+          <li class="nav__item"><a href="#" class="nav__link is-active">Apple</a></li>
+          <li class="nav__item"><a href="#" class="nav__link">Samsung</a></li>
```
anstsot, line 22: add empty lines between multiline sibling blocks of HTML
image
```suggestion
          <li class="nav__item"><a href="#" class="nav__link is-active">Apple</a></li>
          
          <li class="nav__item"><a href="#" class="nav__link">Samsung</a></li>
```

src/styles/blocks/normalize.scss
```diff
@@ -0,0 +1,4 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
```
anstsot, line 3: it's better not to set `margin: 0` for all elements on page. just add this styles for specific selectors
```suggestion
```

src/styles/blocks/is-active.scss
```diff
@@ -0,0 +1,13 @@
+.is-active {
```
anstsot, line 1: this style belongs to nav so it would be better to place it there

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,99 @@
...
+    @extend %font-default;
+  }
+
+  & p {
+    @extend %font-default;
+  }
+
+  & h2 {
+    @extend %font-default;
+  }
```
anstsot, line 18: it's better not to add styles to tags. better to use specific class names (e.g. `&__name`)

src/styles/blocks/is-active.scss
artem-ivashchenko: the test says that .is-active must be a separate class, so it's impossible to implement that way

agniya-i (changes requested):
there are no changes for previous comments, please push them

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,108 @@
...
+  &__code {
+    @extend %font-default;
+
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    color: $card-p-color;
+
+    margin: 0;
+    margin-bottom: 16px;
```
polosanya, line 33: You can use shorthand properties instead
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#syntax

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,108 @@
...
+
+  &__reviews {
+    @extend %font-default;
+
+    margin: 0;
+
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    display: flex;
```
polosanya, line 51: Pay attention to CSS properties order. Check yourself everywhere
https://9elements.com/css-rule-order/

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,108 @@
...
+    text-align: center;
+    text-decoration: none;
+    text-transform: uppercase;
+    color: $color-main;
+    background-color: $color-secondary;
+
+    &:hover {
+      color: $color-secondary;
+      background-color: $color-main;
+      border: 1px solid $color-secondary;
```
polosanya, line 105: It's better to add a transparent border for the default state and just change its color on hover

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,19 @@
...
+  height: 60px;
+  justify-content: space-between;
+  font-size: 12px;
+  text-transform: uppercase;
+  background: $color-main;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  &__logo-pic {
+    width: 40px;
+    height: 40px;
```
polosanya, line 12: You can use SASS mixin here
https://sass-lang.com/documentation/at-rules/mixin

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,18 @@
+.stars {
+  display: flex;
+
+  &__star {
+    height: 16px;
+    width: 16px;
```
polosanya, line 6: Same here

polosanya (changes requested):
Almost perfect!

sTorba24 (approved):
Good work!

PR #2869, opened 2023-03-14
Task solution

SuaBy514:
- [DEMO LINK](https://suaby514.github.io/layout_catalog/)
- [TEST REPORT LINK](https://suaby514.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -113,13 +113,13 @@ however, the linter didn't like that, so I quit that approach */
   font-style: normal;
   text-decoration: none;
   font-family: "Roboto", sans-serif;
-}
 
-/* Hover function */
+  /* Hover function */
 
-.card__button:hover {
-  background: transparent;
-  color: $color-cyan;
-  border: 1px solid $color-cyan;
-  padding: 11px 0;
+  &:hover {
+    background: transparent;
+    color: $color-cyan;
+    border: 1px solid $color-cyan;
+    padding: 11px 0;
+  }
 }
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -1,0 +1,125 @@
+$text-color-primary: #060b35;
+$text-color-secondary: #616070;
+$color-cyan: #00acdc;
+$line-height-small: 14px;
+$line-height-big: 18px;
+$font-size-small: 10px;
+$font-size-big: 12px;
+$font-weight-low: 400;
+$font-weight-mid: 500;
+$font-weight-high: 700;
+
+.card {
+  display: flex;
+  width: 200px;
+  flex-direction: column;
+  margin: 0;
+  margin-left: 48px;
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card:first-child {
+  margin-left: 0;
+}
+
+.card__imac {
+  display: flex;
+  justify-content: center;
+}
+
+.card__imac-image {
+  display: block;
+  position: relative;
+  height: 134px;
+  width: 160px;
+  margin-bottom: 40px;
+}
+
+/* I wanted to group the texts by groups of colors and styles,
+however, the linter didn't like that, so I quit that approach */
+
+.card__text-title {
+  margin: 0;
+  line-height: $line-height-big;
+  font-size: $font-size-big;
+  font-weight: $font-weight-mid;
+  color: $text-color-primary;
+  font-style: normal;
+}
+
+.card__text-sub {
+  margin: 4px 0;
+  line-height: $line-height-small;
+  font-size: $font-size-small;
+  font-weight: $font-weight-low;
+  color: $text-color-secondary;
+  font-style: normal;
+}
+
+.card__reviews,
+.card__price {
+  margin: 12px 0;
+  display: flex;
+  flex-direction: row;
+  width: 166px;
+  justify-content: space-between;
+  align-items: flex-end;
+}
+
+.card__text-review {
+  margin: 0;
+  line-height: $line-height-small;
+  font-size: $font-size-small;
+  font-weight: $font-weight-low;
+  color: $text-color-primary;
+  font-style: normal;
+}
+
+.card__text-price-title {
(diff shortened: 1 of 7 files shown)


kq333 (approved):

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
+@import "./variables.scss";
```
d-wasilewski, line 1: By convention, partial files (containing reusable code) in SCSS are given names that begin with an underscore (e.g. _variables.scss) to indicate that they should not be compiled into a standalone CSS file.

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+}
+
+.card:first-child {
+  margin-left: 0;
+}
+
+.card__imac {
```
d-wasilewski, line 18: checklist point 10, in many places

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+  color: $text-color-secondary;
+  font-style: normal;
+}
+
+.card__text-price-value {
+  margin: 0;
+  line-height: $line-height-big;
+  font-size: 16px;
+  font-weight: $font-weight-high;
+  color: #060b35;
```
d-wasilewski, line 84: should be a variable

src/styles/nav.scss
```diff
@@ -0,0 +1,67 @@
...
+  color: $text-color-primary;
+  font-weight: $font-weight-mid;
+  text-decoration: none;
+  margin-right: 0;
+  margin-left: 20px;
+  height: 60px;
+  /* Readme said to not use uneven numbers, this is a workaround
+     And despite what the checklist says, the distance between the
+     content on Figma IS 22.5 */
+  padding: calc(45px / 2) 0;
```
d-wasilewski, line 43: Try to center it with flexbox

src/styles/row.scss
```diff
@@ -0,0 +1,10 @@
+.row {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin-bottom: 46px;
+}
```
d-wasilewski, line 6: We want to do somethinig similar to grid here, so if we have 7 cards or less they should be on the same level. Relates to checklist 13
image

d-wasilewski: Also, the header should have border on the bottom 😄

d-wasilewski (changes requested):

src/styles/nav.scss
SuaBy514: The disadvantage of using that is the clickable area becomes only the text itself, making it more difficult to click on the nav

src/styles/row.scss
SuaBy514: While I understand that, the readme states "there should always be 4 cards in a row (not 2, 3 or 5)"

SuaBy514: The header also does not have a border in the Figma mock-up we were given

d-wasilewski: @SuaBy514 sorry, I meant the shadow 
image

src/styles/nav.scss
d-wasilewski: It actually stays the same, so you dont have to worry about that
image

src/styles/main.scss
```diff
@@ -1,3 +1,33 @@
...
 }
+
+/* For some reason, using HTML to load the SCSS files
+   was suddenly causing mate scripts to crash, this is
+   an alternative which doesn't cause crashes */
+
+@import "./nav.scss";
+@import "./stars.scss";
+@import "./card.scss";
+@import "./row.scss";
```
d-wasilewski, line 33: Imports should always be at the beggining of the file

src/styles/row.scss
```diff
@@ -0,0 +1,10 @@
+.row {
+  display: flex;
+  align-items: flex-start;
+  width: 944px;
```
d-wasilewski, line 4: checklist 13, create a function

src/styles/card.scss
```diff
@@ -0,0 +1,113 @@
...
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &:first-child {
+    margin-left: 0;
+  }
+}
+
+.card__imac {
```
d-wasilewski, line 18: Make use of the sass nesting, this should be `&__imac`, fix it everywhere

d-wasilewski (changes requested):

src/styles/row.scss
d-wasilewski: I think the text in readme meant that it always need to be placed in the column of the same amount of cards, not different. So this I stated behaviour should be preserved

src/styles/nav.scss
SuaBy514: Then I think I'm doing it wrong, because when I do it it ends up reducing the clickable space to just the text
Could I get some pointers on how I should go about this?

src/styles/row.scss
SuaBy514: I see, I'll see what I can do about the grid thing without actually using a grid object (since that's an assignment that's still ahead), but I expect it to take me a while

src/styles/card.scss
SuaBy514: I.. honestly did not know it could be used for object children as well, every example just used it with functions like :hover, so I assumed that's what that's for
My bad, I will make use of it properly from now on

src/styles/stars.scss
```diff
@@ -0,0 +1,26 @@
+@import "./_variables.scss";
+
+.stars {
+  list-style: none;
+  display: flex;
+  @include delete-margin-padding();
+}
+
+.stars__star {
```
michalcywka, line 9: use `&`

michalcywka (changes requested):
I've read the comments, I will accept the PR just make this one small change I marked.

michalcywka (approved):

PR #2871, opened 2023-03-14
Develop

Dka888:
- [DEMO LINK](https://Dka888.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Dka888.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/parts/card.scss
+++ b/src/styles/parts/card.scss
@@ -7,7 +7,7 @@
   background-color: #fff;
   padding: 32px 16px 16px;
   margin-right: 48px;
-  margin-bottom: 48px;
+  margin-bottom: 46px;
 
   &__apple {
     display: block;
@@ -30,10 +30,11 @@
     display: flex;
     justify-content: space-between;
     align-items: center;
-    margin-bottom: 24px;
+    margin-bottom: 23px;
 
     &-number {
       font-size: 10px;
+      line-height: 14px;
       font-weight: 400;
       color: $color-main-accent;
       font-family: "Roboto", sans-serif;
@@ -75,7 +76,7 @@
     height: 40px;
 
     box-sizing: border-box;
-    border: 1px transparent;
+    border: 1px solid $color-accent-blue;
     border-radius: 5px;
 
     text-decoration: none;
@@ -92,9 +93,8 @@
     font-weight: 700;
 
     &:hover {
-      background-color: white;
+      background-color: #fff;
       color: $color-accent-blue;
-      border: 1px solid $color-accent-blue;
     }
   }
 
--- a/src/styles/parts/card.scss
+++ b/src/styles/parts/card.scss
@@ -1,0 +1,109 @@
+@import "../utils/_variables.scss";
+
+.card {
+  width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  background-color: #fff;
+  padding: 32px 16px 16px;
+  margin-right: 48px;
+  margin-bottom: 46px;
+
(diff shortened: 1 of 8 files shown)


bjeziorek (changes requested):

src/styles/parts/card.scss
```diff
@@ -0,0 +1,104 @@
+.card {
+  width: 200px;
+  border: 1px solid $color-white-color;
+  border-radius: 5px;
+  background-color: #fff;
+  padding: 32px 16px 16px;
+}
+
+.card__apple {
```
bjeziorek, line 9: You don't use sass syntax. It should be
```
.card {
  &__apple{
  //
  }
} 
```
instead of css syntax:
```
.card {}
.card__apple {}
```
and so on in card.scss file. Other files look fine.

src/styles/parts/card.scss
```diff
@@ -0,0 +1,108 @@
+@import "../utils/_variables.scss";
+
+.card {
+  width: 200px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  background-color: #fff;
+  padding: 32px 16px 16px;
+  margin-right: 48px;
+  margin-bottom: 46px;
```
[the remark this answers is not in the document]
bjeziorek: You shouldn't use .card selector as it's BEM block. You need to add extra class which will be container's element like .cardContainer which includes .card and the html element with .card should have also .cardContainer__element class (name you as you wish, just follow container's name), only then BEM allows to set margin on BEM element (like .cardClass__element), on block (like .card) it's disallowed. It will work in the same way and it's more work to do but that's BEM rule.

bjeziorek (changes requested):

src/styles/parts/card.scss
```diff
@@ -0,0 +1,104 @@
...
+    &:hover {
+      background-color: $color-white-color;
+      border-color: $color-accent-blue;
+      color: $color-accent-blue;
+    }
+  }
+
+  &:nth-child(-n+#{$cards-number * 2}) {
+    margin-right: 48px;
+    margin-bottom: 46px;
```
bjeziorek, line 98: I think that it still counts as adding margin on BEM block and should be moved to any BEM element of container like for example page__card, just can't be on BEM block.

src/styles/parts/nav.scss
```diff
@@ -0,0 +1,42 @@
+@import "../utils/_variables.scss";
+
+.nav {
+  padding: 0;
+  margin: 0;
```
bjeziorek, line 5: The same BEM issue - margin can't be on BEM block. .nav is BEM block. You have header above nav, you can use it as block (.header is BEM block, .nav <-- in this form it's also BEM block, where you can't style margins, even if it's physically possible and work on browser) and place on its child class .header__nav (<- it's BEM element where you can style margins ) and move all margins from .nav{} to .header__nav {} and your nav should be like this <nav class="nav header__nav>. And please check all your classes, I see margins set in many places of your BEM blocks also in other files.

bjeziorek (changes requested):
Everything else looks fine, just try to do something with these margins on BEM block (I know it makes a mess, and why to change something that works - because of BEM rules, I have the same issue so I know how frustrating it can be ;), but I really can't approve... ).

bjeziorek (approved):
You still have disallowed styles in BEM blocks like margins but I was told to let it pass so I do.

michalcywka (changes requested):
Checklist point 2

michalcywka (approved):

PR #2881, opened 2023-03-15
layout_catalog to check

PufAnimalsDev:
- [DEMO LINK](https://pufanimalsdev.github.io/layout_catalog/)
- [TEST REPORT LINK](https://pufanimalsdev.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,9 +4,371 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet"> 
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Catalog</h1>
+  <body class="page-content">
+    <header class="header page-content__header">
+      <div class="header__logo-wrapper">
+        <img 
+          src="./images/logo.png" 
+          alt="Moyo logo" 
+          class="header__logo"
+        >
+      </div>
+      <nav>
+        <ul class="header__menu">
+          <li class="header__menu-item">
+            <a 
+              href="#apple" 
+              class="header__menu-link is-active"
+            >Apple</a>
+          </li>
+
+          <li class="header__menu-item">
+            <a href="#Samsung" class="header__menu-link">Samsung</a>
+          </li>
+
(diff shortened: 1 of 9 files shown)


PufAnimalsDev:
> Requirements: there should always be 4 cards in a row (not 2, 3 or 5)

can you tell me more? The point that when I shrinking the screen, my cards jump to a new column?

PufAnimalsDev:
> Requirements: there should always be 4 cards in a row (not 2, 3 or 5)

if I understood correctly, I didn't decide to change it because, I would have to give a const width to block main, and I would like to give width only to the element in which I have cards (section with content), if I want added another section to main, maybe it would have a different width, and I don't think responsiveness was required for this task.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,63 @@
...
+    height: 4rem;
+    width: 4rem;
+    margin: 1rem 0;
+  }
+
+  &__menu {
+    display: flex;
+    list-style: none;
+    margin: 0;
+    padding: 0;
```
d-wasilewski, line 22: this should be handled in _reset.scss

d-wasilewski (changes requested):
G

src/index.html
```diff
@@ -4,9 +4,371 @@
...
+            class="product-card__image" 
+            alt="APPLE A1419 iMac 27 Retina 5K"
+          >
+          <h2 class="product-card__title">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h2>
+          <p class="product-card__code">
+            Product code: 195434
+          </p>
+          <div class="product-card__reviews-wrapper">
```
d-wasilewski, line 80: The same thing I mentioned earlier - remember to scan your code for similar mistakes after each review
```suggestion
          <div class="product-card__reviews">
```

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,94 @@
...
+    align-items: center;
+    text-decoration: none;
+    text-transform: uppercase;
+    font-size: 1.4rem;
+    font-weight: 700;
+    border-radius: $card-border-radius;
+    border: solid 0.1rem transparent;
+    line-height: 1.6rem;
+    padding: 1.1rem 0;
+    transition: all 0.3s ease-in-out;
```
d-wasilewski, line 86: be more explicit
```suggestion
    transition: color 0.3s ease-in-out, another-property 0.3s ease-in-out etc...;
```

d-wasilewski (changes requested):

PR #2884, opened 2023-03-15
Preliminary version

Ozwiena:
- [DEMO LINK](https://ozwiena.github.io/layout_catalog/)
- [TEST REPORT LINK](https://ozwiena.github.io/layout_catalog/report/html_report/)

I know it does not work (only 6 test out of 10 are passed). I will work on the task tomorrow.


michalcywka (changes requested):
We don't need `dist` folder to be commited here :)

michalcywka (approved):

PR #2886, opened 2023-03-15
done

Timur-Rudenko:
https://Timur-Rudenko.github.io/layout_catalog/
https://Timur-Rudenko.github.io/layout_catalog/report/html_report/


loralevitska (changes requested):
Keep up the good work! Some comments to fix:
Try to pass all failed tests. Pay attention for indents. Your version more lefter than need
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

PR #2900, opened 2023-03-16
firstEdition

oleksandr-rudiuk:


src/index.html
```diff
@@ -4,9 +4,421 @@
...
+
+    <main class="main">
+      <section class="main__container">
+        <div
+          class="card main__item"
+          data-qa="card"
+        >
+          <img class="card__image"
+          src="./images/imac.jpeg"
+          alt="Apple computer picture"
```
OleziO, line 102: Each property must start from a new line with 2 spaces more than the tag. Check other similar cases.

src/styles/card.scss
```diff
@@ -0,0 +1,123 @@
...
+  height: 16px;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--1:first-child {
+  margin-left: 16px;
+}
+
+.stars--1 :nth-child(-n+4) {
```
OleziO, line 66: Add code for all cases (from 1 to 5 stars)

OleziO (changes requested):
image
image

SanyaBratashchuk (changes requested):
image

src/styles/main.scss
```diff
@@ -1,3 +1,52 @@
...
+.main {
+  display: flex;
+  justify-content: center;
+
+  &__container {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    padding: 50px 40px;
+    align-self: center;
```
TarasHoliuk, line 30: would be nice to add some `width` / `max-width` for the container
it looks not good on wider screens:
image

But since it's not required in the task - not critical and not necessary to fix it

src/styles/main.scss
```diff
@@ -1,3 +1,52 @@
...
+  .main__container {
+    padding: 50px 0;
+  }
+}
+
+@media (min-width: 1200px) {
+  .main__container {
+    padding: 50px 128px;
+  }
+}
```
TarasHoliuk, line 52: Not sure you need it, anyway:
you can write media queries directly in `&__container` selector, for example:

```
&__container {
  display: flex;
  flex-direction: row;
  flex-wrap: wrap;
  padding: 50px 40px;
  align-self: center;
    
  @media (min-width: 1200px) {
    padding: 50px 128px;
  }
}
```
Just FYI

TarasHoliuk (approved):
Approved with 2 comments (just FYI - no need to re-request review)

PR #2914, opened 2023-03-22
added task solution

mariuszgit:
https://mariuszgit.github.io/layout_catalog/
https://mariuszgit.github.io/layout_catalog/report/html_report/


src/styles/blocks/_main-content.scss
```diff
@@ -0,0 +1,26 @@
...
+.main-content {
+  display: flex;
+  flex-wrap: wrap;
+  padding: 50px 40px;
+  width: 1024px;
+
+  @include large-up {
+    padding: 50px 128px;
+    width: 1200px;
+  }
```
DorotaLeniecDev, line 12: This part is not needed, as with width growth you also raise left/right padding so you can stay with default styles for this main-content

src/styles/blocks/_main-content.scss
```diff
@@ -0,0 +1,26 @@
...
+    margin-right: 48px;
+  }
+
+  &__item:nth-child(4n) {
+    margin-right: 0;
+  }
+
+  &__item:nth-last-child(-n + 4) {
+    margin-bottom: 0;
+  }
```
DorotaLeniecDev, line 25: You can use nesting and add those styles inside **&__item{}**

src/styles/blocks/_product-card.scss
```diff
@@ -0,0 +1,83 @@
+@use "../utils/variables" as *;
+
+.product-card {
+  background-color: #fff;
```
DorotaLeniecDev, line 4: You can create variable for this color as it is used more than once

src/styles/blocks/_product-card.scss
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: 32px 16px 16px;
+  width: 200px;
+
+  &__image {
+    align-self: center;
+    height: 134px;
+    margin-bottom: 40px;
+    vertical-align: bottom;
```
DorotaLeniecDev, line 16: Is this neccessary?

src/styles/blocks/_stars.scss
```diff
@@ -0,0 +1,21 @@
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 100px;
+  /*dont undersand why need this padding, but its required to pass tests*/
+  padding-right: 4px;
```
DorotaLeniecDev, line 6: I think better option would be to add margin-right to each star element, rather than adding fixed width.

DorotaLeniecDev (changes requested):
Good job! Few adjustments needed, check my comments :)

PR #2923, opened 2023-03-23
create catalog

pos1bl:
- [DEMO LINK](https://pos1bl.github.io/layout_catalog/)
- [TEST REPORT LINK](https://pos1bl.github.io/layout_catalog/report/html_report/)

According to the 14th condition in CODE STYLE, I created a function that calculates the width of the directory. (1 screenshot) I also added variables for the number of columns and the length of the cards. (2 screenshots) This way I made sure that our catalog was flexible and could easily adapt if we want to change the number of columns in the catalog, for example, to any number. Also, we can easily increase the width of the card in a variable and the page will look good.

I also deviated a little from the condition and wrote the code (3 screenshots) so that we could easily add items to the catalog and our page adapted to this, and all elements had indents from above, except for the elements of the first line.

image
image
image

### DIFF
--- a/src/styles/nav.scss
+++ b/src/styles/nav.scss
@@ -1,0 +1,36 @@
+.nav {
+  height: 60px;
+  position: relative;
+  display: inherit;
+
+  &__list {
+    display: inherit;
+    text-transform: uppercase;
+    padding: 0;
+    justify-content: flex-end;
+    margin: 0;
+  }
+  
+  &__item {
+    margin-left: 20px;
+    display: inherit;
+
+    &:first-child {
+      margin-left: 0;
+    }
+  }
+  
+  &__link {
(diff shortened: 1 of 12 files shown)


src/styles/nav.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+  
+  &__item {
+    margin-left: 20px;
+    display: inherit;
+
+    &:first-child {
+      margin-left: 0;
+    }
+  }
```
etojeDenys, line 21:
```suggestion
  &__item:not(:first-child) {
    margin-left: 20px;
  }
```

etojeDenys (changes requested):
almost done

vlad-demchuk (approved):
GJ

PR #2941, opened 2023-03-31
add task solution

4ykis:
https://4ykis.github.io/layout_catalog/
https://4ykis.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/components/_card-container.scss
+++ b/src/styles/components/_card-container.scss
@@ -1,0 +1,20 @@
+.card-container {
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: wrap;
+  max-width: maxWidth(4, 48);
+  margin: 0 auto;
+  padding: 50px 0;
+
+  &__item {
+    margin: 0 48px 46px 0;
+
+    &:nth-child(4n) {
+      margin-right: 0;
+    }
+
+    &:nth-last-child(-n + 4) {
+      margin-bottom: 0;
+    }
+  }
+}
--- a/src/styles/components/_card-container.scss
+++ b/src/styles/components/_card-container.scss
@@ -1,6 +1,6 @@
 .card-container {
   display: flex;
-  justify-content: space-between;
+  justify-content: flex-start;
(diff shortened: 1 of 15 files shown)


src/styles/components/_card.scss
```diff
@@ -0,0 +1,69 @@
+@import "./card/price";
+@import "./card/stars";
+@import "./card/reviews";
+
+.card {
+  box-sizing: border-box;
+  font-weight: 400;
```
YegorKorenkov, line 7: font-weight 400 by default, redundant

src/styles/components/_card.scss
```diff
@@ -0,0 +1,69 @@
+@import "./card/price";
+@import "./card/stars";
+@import "./card/reviews";
+
+.card {
+  box-sizing: border-box;
```
YegorKorenkov, line 6: please move to * selector

src/styles/components/_card.scss
```diff
@@ -0,0 +1,69 @@
...
+    margin-bottom: 16px;
+  }
+
+  &__button {
+    box-sizing: border-box;
+    display: block;
+    border-radius: 5px;
+    background-color: $color-blue;
+    border: 1px solid $color-blue;
+    line-height: 38px;
```
YegorKorenkov, line 55: Please use vertical paddings instead line height

src/styles/components/_card-container.scss
```diff
@@ -0,0 +1,20 @@
+.card-container {
+  display: flex;
+  justify-content: space-between;
```
YegorKorenkov, line 3: More correctly use justify-content: center, because space-between can set additional indents and they are already set with

YegorKorenkov (changes requested):

src/styles/components/_card-container.scss
4ykis: agreed, but than `flex-start` looks better.

VitaliyBondarenko1982 (approved):

PR #2942, opened 2023-03-31
ready!!

danchkk:
- [DEMO LINK](https://danchkk.github.io/layout_catalog)
- [TEST REPORT LINK](https://danchkk.github.io/layout_catalog/report/html_report/)


etojeDenys (changes requested):
1. let's create separate scss files for blocks (e.g: `page`, `stars`, `card`)
2.   Screen Recording 2023-04-01 at 10.30.14.mov

etojeDenys (changes requested):
I still can see the hover effect only on the first card

danchkk:
> I still can see the hover effect only on the first card

i fixed it in commit 3 but for some reason the fix is ​​not visible on the DEMO LINK

VitaliyBondarenko1982 (approved):
see hover effect in other cards

PR #2947, opened 2023-03-31
Develop

GlibVK:
https://glibvk.github.io/layout_catalog/
https://glibvk.github.io/layout_catalog/report/html_report/


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,126 @@
+@use "../utils/variables" as *;
+
+.main {
+  display: flex;
+  justify-content: center;
+}
+
+.catalog {
+  margin-top: 50px;
```
Volodymyr-Melnyk-Fron10der, line 9: do not set margin and position for block , use mixes

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,126 @@
...
+    margin-right: 48px;
+    margin-bottom: 46px;
+
+    &:nth-child(4n) {
+      margin-right: 0;
+    }
+  }
+}
+
+.card {
```
Volodymyr-Melnyk-Fron10der, line 26: move block into separated file

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,126 @@
+@use "../utils/variables" as *;
+
+.main {
```
Volodymyr-Melnyk-Fron10der, line 3: looks like separated block, I would move into a file

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,68 @@
...
+    font-family: Roboto, sans-serif;
+    font-style: normal;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 14px;
+    text-transform: uppercase;
+    text-decoration: none;
+    color: $color-dark-blue;
+
+    height: 60px;
```
Volodymyr-Melnyk-Fron10der, line 46: do not hardcode the height, use paddings

Volodymyr-Melnyk-Fron10der (changes requested):
1. layout shouldn't be broken if I add some more cards
image

PR #2973, opened 2023-04-10
added product cards to the page, but does not skip 2 tests

rkhramiuk:
- [DEMO LINK](https://rkhramiuk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://rkhramiuk.github.io/layout_catalog/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
-body {
-  margin: 0;
-}
+@import "./utils/variables";
+@import "./utils/normalize.scss";
+@import "./scss-folder/header.scss";
+@import "./scss-folder/nav.scss";
+@import "./scss-folder/product-card.scss";
+@import "./scss-folder/stars.scss";
```
loralevitska, line 6: add this style files to git for review

loralevitska (changes requested):
Good job! There are things that should be fixed:
Try to fix 2 failed tests 
The 1st screenshot is what the padding should look like and the 2nd screenshot is what your solution looks like, that's why the tests fail because you used the wrong padding for the container
image
image

etojeDenys (approved):
awesome

VitaliyBodnar (approved):

PR #2977, opened 2023-04-11
Develop

Dolnys:
[DEMO LINK](https://Dolnys.github.io/layout_catalog/)
[TEST REPORT LINK](https://Dolnys.github.io/layout_catalog/report/html_report/)


anpiet:
- there should always be 4 cards in a row (not 2, 3 or 5). When width goes below 1024px, there is less cards in rows.

anpiet (changes requested):
there should always be 4 cards in a row (not 2, 3 or 5). When width goes below 1024px, there is less cards in rows.

Dolnys:
> there should always be 4 cards in a row (not 2, 3 or 5). When width goes below 1024px, there is less cards in rows.

 Good point, I added the min width of the content to 1024 px so now there should always be 4 elements in a row.

anpiet (approved):

src/index.html
```diff
@@ -1,12 +1,291 @@
...
+          <span class="card__price-span">Price:</span>
+          <span class="card__price-amount">$2,199</span>
+        </div>
+        <a
+        href="#"
+        class="card__button-link"
+        data-qa="card-hover">Buy
+        </a>
+      </div>
+      <div class="card">
```
mvjl000, line 107: Checklist point 1

I would add empty line between cards, your html would look way cleaner ;)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,7 @@
+$color-main-accent: #060b35;
+$color-secondary: #616070;
+$color-accent-blue: #00acdc;
+$color-border-card: #f3f3f3;
+$img-product-height: 134px;
+$img-product-width: 160px;
+$img-star-height-width: 16px;
```
mvjl000, line 7: Checklist point 11

Create variables only for values that are being used more than once

src/styles/blocks/content_container.scss
```diff
@@ -0,0 +1,9 @@
+.main {
+  display: flex;
+  min-width: 1024px;
+  max-width: 1200px;
+  justify-content: center;
+  flex-wrap: wrap;
+  padding: 50px 40px;
+  margin: 0 auto;
+}
```
mvjl000, line 9: It is said in the readme that if you are familiar with grid it is mandatory to use it.
I'm aware of that you have done this task a while ago but I would like you to refactor your layout to grid. And additionally add media queries for [RWD](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Responsive_Design) - remember about mobile first approach.
Here are the [grid layout instrutions](https://github.com/mate-academy/layout_catalog/blob/master/grid.md) from the readme

src/styles/blocks/navigation.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  &__item-link:link,
+  &__item-link:visited {
+    color: $color-main-accent;
+    text-decoration: none;
+    font-weight: 500;
+    display: flex;
+    align-items: center;
+    transition: all 0.3s;
```
mvjl000, line 40: List all the properties that you want to have transition.

mvjl000 (changes requested):

src/styles/blocks/navigation.scss
mvjl000: @Dolnys one thing I forgot about - add smooth transitions and animations. Check the readme

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,108 @@
...
+  display: flex;
+  flex-direction: column;
+  width: 200px;
+  padding: 32px 16px 16px;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  background-color: #ffff;
+  transition: transform 0.3s;
+  // margin-right: 48px;
+  // margin-bottom: 46px;
```
choeqq, line 13: please remove comments before pushing commit

src/styles/blocks/content_container.scss
choeqq, line 1: I am thinking of changing file name to `main` 🤔 - typically we use same file name as BEM block - I think it would be better to keep it like in other files, beacuse technically it is `content_container` but we've agreed upon file structure convention - so I wouldn't really change that 😉

choeqq (changes requested):
Code looks really good, but some tests are still failling due to size of cards differences, please try to fix it and you're good to go 🚀

src/styles/blocks/card.scss
Dolnys: Done

src/styles/blocks/content_container.scss
Dolnys: I changed the 'main' class to 'card', 'card' to 'card__content', and removed whole content__container file

Dolnys:
I have deleted the comments and changed the file name, but I don't know how to get rid of these errors. I have spent several hours on it, and it still doesn't pass the tests. After consulting with Anastasia, I should leave the code as it is.

PR #2991, opened 2023-04-16
add task solution

maksym-nemera:
- [DEMO LINK](https://maksym-nemera.github.io/layout_catalog/)
- [TEST REPORT LINK](https://maksym-nemera.github.io/layout_catalog/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,115 @@
+.card {
+  width: $widthCard; // 200px
```
naziamoff, line 2: no need for the comments here.
Imagine the page was redesigned - what will you do? You will need to fix all the comments manually. This way variable makes no sense if you need to fix it manually in the code 😄 
```suggestion
  width: $widthCard;
```

src/index.html
```diff
@@ -5,8 +5,487 @@
...
+              Video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="main__cards">
+        <div class="card card--margin-right-bottom" data-qa="card">
```
naziamoff, line 89: and this is list item

src/index.html
```diff
@@ -5,8 +5,487 @@
...
+              <h5 class="card__cost">
+                $2,199
+              </h5>
+            </div>
+
+            <a
+            href="#"
+            data-qa="card-hover"
+            class="card__button"
+            >
```
naziamoff, line 133:
```suggestion
            <a
              href="#"
              data-qa="card-hover"
              class="card__button"
            >
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,118 @@
...
+  box-sizing: border-box;
+
+  background: $background-color;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &--margin-right-bottom {
+    margin-right: $marginRightCard;
+    margin-bottom: 46px;
+  }
```
naziamoff, line 14: that is called [atomic css](https://css-tricks.com/lets-define-exactly-atomic-css/)
I would say you should not add classes like that, because from them I don't understand what exactly piece of code is that

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,118 @@
...
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
+
+  background: $background-color;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &--margin-right-bottom {
+    margin-right: $marginRightCard;
```
naziamoff, line 12: it is good creating variables, but variable for margin is killing a bee with a sledgehammer 🐝 🔨 = 💀 🔨

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,118 @@
...
+    display: flex;
+    justify-content: space-between;
+
+    &--flex-b24 {
+      margin-bottom: 24px;
+    }
+
+    &--flex-b16 {
+      margin-bottom: 16px;
+    }
```
naziamoff, line 60: same here, it is called atomic. You can add some atomic styles if you want, but this can't be inside bem block, you just create separate styles for that, e.g.
```
.mb-16 {
  margin-bottom: 16px
}
```
And just add this as a class to the element where you need it

Or find some lib which does that

naziamoff (changes requested):
that's impressive job, but we want it even closer to perfect. So see details in comments

src/styles/blocks/card.scss
NickMitchell61: In creating this class I try to avoid changing the main class "card" by margins. I've tried to save natural borders in the card.

I'll have deleted it)

src/styles/blocks/card.scss
```diff
@@ -8,10 +8,8 @@
   border: 1px solid #f3f3f3;
   border-radius: 5px;
 
-  &--margin-right-bottom {
-    margin-right: $marginRightCard;
-    margin-bottom: 46px;
-  }
+  margin-right: 48px;
+  margin-bottom: 46px;
```
naziamoff, line 12: I would actually create some wrapper containing your block for that, for not influencing the outer environment

naziamoff (approved):
nice 🔥

PR #3021, opened 2023-04-24
catalog solution

PANCH1811:
- [DEMO LINK](https://panch1811.github.io/layout_catalog/)
- [TEST REPORT LINK](https://panch1811.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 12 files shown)


PANCH1811:
idk why i`m failing that 2 tests, "reference" and "test" looks same, completely no difference

Shevtsooov (changes requested):
Hey! Well done!
However, I would add some semantics to your code)

It is strange that your code hasn't passed the test, as the `main` tag is present in the html :(

silchyn (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,90 @@
+.card {
+  width: $card-width;
+  padding: 16px;
+  background: $color-main-background;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
```
bohdan-mykhailenko, line 6: It's better to group similar style values(e.g. font styles, position properties).

src/styles/blocks/page-cards.scss
```diff
@@ -0,0 +1,19 @@
+.page-cards {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  width: PageWidth($card-width, $card-margin-right, $cards-per-row);
```
bohdan-mykhailenko, line 5: Here you got the wrong value of width because you needn't use the margin-right value of last element in row.

src/styles/utils/normalize.scss
```diff
@@ -0,0 +1,10 @@
+@import url(https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700;900&display=swap);
```
bohdan-mykhailenko, line 1: Maybe it's a better way to use a tag 'link' in the tag 'head' for fonts rather than import it in the scss file.

bohdan-mykhailenko (approved):
Good job 👍  Here are some thoughts to make it perfect

src/styles/blocks/page-cards.scss
PANCH1811: @bohdan-mykhailenko what u mean wrong? can't see the mistake, pls advise.

src/styles/utils/normalize.scss
PANCH1811: @bohdan-mykhailenko why u think so? Actually, maybe it's a bit better to link a font to html tag performance vise, but it really isn’t worth worrying too much about  in this case.

src/index.html
```diff
@@ -7,6 +7,325 @@
...
+            <a class="nav__link" href="#photo">photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#video">video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="page-cards">
```
silchyn, line 60:
```suggestion
    </header>
    
    <main class="page-cards">
```

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,61 @@
...
+
+.moyo-logo {
+  align-self: center;
+  left: 50px;
+
+  width: 40px;
+  height: 40px;
+}
+
+.nav {
```
silchyn, line 20: `nav` and `moyo-logo` are not part of `header` block

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,3 @@
+@function PageWidth($width, $margin, $quantity) {
+  @return $width * $quantity + $margin * $quantity;
```
silchyn, line 2: but there are 3 margins for 4 cards

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin star-size($size) {
```
silchyn, line 1:
```suggestion
@mixin square-size($size) {
```

silchyn (approved):

PR #3024, opened 2023-04-24
task layout catalog

msukhovych:
demo https://msukhovych.github.io/layout_catalog/
test https://msukhovych.github.io/layout_catalog/report/html_report/


qoa11a (changes requested):
Nice code style! 👍 
But you have some styles errors:
- Make `.cards-container` with `max-width: 1024px;`, then just centre it with `margin: 0 auto;`.
- Give all cards `.card:not(:nth-child(4n+1)) {
    margin-left: 48px;
}` and remove other margins.

Start with these errors and continue towards fixing other styles according to these changes. Good luck! 🍀 

Thread in Slack: https://mate-parttime.slack.com/archives/C041JA5EJKV/p1682359460726259

polosanya (changes requested):
Vertical margin between cards should be 46px(not 46), as the design doesn't include borders
image

src/styles/blocks/cards-catalog.scss
```diff
@@ -0,0 +1,113 @@
+.cards-catalog {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+
+  &:nth-child(-n+4) {
+    margin-top: 49px;
```
polosanya, line 7: This margin should be 50px instead of 49 because the borders are not included in the design

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,37 @@
+.card__rating {
```
polosanya, line 1: It's an element of `card` block, so it shouldn't be in this file. Move into the file with card styles

polosanya (changes requested):
A few more changes

OleziO (changes requested):
Fix items aligning. Item must start from the start and don't break others.
image

PR #3040, opened 2023-04-26
add task solution

LizKumber:
https://LizKumber.github.io/layout_catalog/
https://LizKumber.github.io/layout_catalog/report/html_report/


VitaliyBondarenko1982 (changes requested):
please check demo links. they don't work
maybe try to deploy again.

PR #3071, opened 2023-04-27
add task solution

LeonIllya:
- [DEMO LINK](https://LeonIllya.github.io/layout_catalog/)
- [TEST REPORT LINK](https://LeonIllya.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,9 +4,394 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a href="#">
+        <img 
+          src="./images/logo.png" 
+          alt="MOYO logo"
+          class="logo"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#apple">apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#samsung">samsung</a>
+          </li>
+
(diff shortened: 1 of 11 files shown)


src/index.html
```diff
@@ -4,9 +4,394 @@
...
+          </li>
+        </ul>
+
+      </nav>
+    </header>
+
+    <main class="page">
+      <section class="catalog page__catalog">
+        <article class="card catalog__catalog-item" data-qa="card">
+          <div>
```
Anna-Tkachenko, line 67: Do you need extra div for the card image? Let's get rid of it

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,6 @@
+// .page {
```
Anna-Tkachenko, line 1: If you don't need these styles - will be better to delete them

src/styles/utils/header.scss
```diff
@@ -0,0 +1,13 @@
...
+  display: flex;
+
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.5);
+}
+
+.logo {
```
Anna-Tkachenko, line 11: Is it the separate BEM block? If yes - it should be in separate scss file

Anna-Tkachenko (changes requested):
Check your styles for header (check tests)
Looks like you've added extra redundant border and font styles doesn't look similar to the tests
Check how you manage it in `Moyo Header` task and just move those header in `catalog` one

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+  max-width: $card-width;
+  padding: 32px 16px 16px;
+  border: 1px solid $border-color;
+  border-radius: 5px;
+
+  &__img {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin-bottom: 40px;
```
Anna-Tkachenko, line 13: Image for the card should be centered

Anna-Tkachenko (approved):
Good job! 
Just check one more time tests width header - looks like you didn't move all correct styles from the previous task

PR #3076, opened 2023-04-28
add task solution

sasdaiv:
- [DEMO LINK](https://sasdaiv.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sasdaiv.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/cards.scss
+++ b/src/styles/blocks/cards.scss
@@ -1,0 +1,91 @@
+.card {
+  box-sizing: border-box;
+  padding: 32px 16px 16px;
+
+  width: $card-width;
+  background: #fff;
+  border: 1px solid $card-border-color;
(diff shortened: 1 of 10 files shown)


src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,91 @@
+.card {
+  box-sizing: border-box;
```
OleziO, line 2: Move to * selector and remove it from all other places.

OleziO (changes requested):

src/styles/blocks/cards.scss
sasdaiv: fixed

nazarmatsevych (approved):

PR #3087, opened 2023-05-02
add task solution

Huracann:
- [DEMO LINK](https://Huracann.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Huracann.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,111 @@
+.card {
+  width: 200px;
+  padding: 32px 16px 16px;
+
+  text-align: center;
+
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &__image {
+    display: block;
+
+    margin: 0 auto 40px;
+
+    width: 160px;
+    height: 134px;
+  }
+
+  &__item-name {
+    width: 166px;
+    margin-bottom: 4px;
+
+    font-size: 12px;
+    line-height: 18px;
+    text-align: start;
+    font-weight: 500;
+    color: $main-text-color;
+  }
+
+  &__product-code {
+    margin-bottom: 16px;
+
+    text-align: start;
+    font-size: 10px;
+    line-height: 14px;
+    color: $secondary-text-color;
+  }
+
+  &__reviews {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    margin-bottom: 23px;
+
+    font-size: 10px;
+    font-weight: 400;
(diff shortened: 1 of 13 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,111 @@
...
+    text-align: start;
+    font-size: 10px;
+    line-height: 14px;
+    color: $secondary-text-color;
+  }
+
+  &__reviews {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
```
naziamoff, line 43: seems you use it more than once, let's create a mixing for that

src/styles/utils/functions.scss
```diff
@@ -0,0 +1,3 @@
+@function dynamicMaxWidth($width, $margins, $amount) {
+  @return $width * $amount + $margins * ($amount - 1);
+}
```
naziamoff, line 3: I would name this function `getMaxWidth` or something. It will be maximally clear this way

src/styles/main.scss
```diff
@@ -1,3 +1,12 @@
...
+
+@import "./utils/reset.scss";
+@import "./utils/normalize.scss";
+
+@import "./blocks/header.scss";
+@import "./blocks/nav.scss";
+@import "./blocks/container.scss";
+@import "./blocks/catalog.scss";
+@import "./blocks/card.scss";
+@import "./blocks/stars.scss";
```
naziamoff, line 12:
```suggestion
@import "./utils/variables.scss";
@import "./utils/functions.scss";
@import "./utils/reset.scss";
@import "./utils/normalize.scss";
@import "./blocks/header.scss";
@import "./blocks/nav.scss";
@import "./blocks/container.scss";
@import "./blocks/catalog.scss";
@import "./blocks/card.scss";
@import "./blocks/stars.scss";
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,111 @@
...
+    color: $button-text-color;
+
+    background: $button-bg-color;
+    border: 1px solid transparent;
+    border-radius: 5px;
+
+    &:hover {
+      color: $button-bg-color;
+      background: white;
+      border-color: $button-bg-color;
```
naziamoff, line 95: It works only for some small project which won't need a lot of support further. What if we gonna need the same color in some other place, not for a button? Will this name be a good choice?

naziamoff (changes requested):
already good, but let's fix the comments and add some transitions across the project

PR #3104, opened 2023-05-06
Develop

RodionSav:


src/styles/main.scss
```diff
@@ -1,3 +1,34 @@
-body {
+$main-color: #00acdc;
+$color-link: #060b35;
+$code-color: #616070;
+$border-color: #f3f3f3;
+$white-color: #fff;
```
Volodymyr-Melnyk-Fron10der, line 5: move into the variables.scss

src/styles/main.scss
```diff
@@ -1,3 +1,34 @@
...
+  padding: 0;
+}
+
+ul {
+  list-style: none;
+}
+
+a {
+  text-decoration: none;
 }
```
Volodymyr-Melnyk-Fron10der, line 31: move into the reset.scss

Volodymyr-Melnyk-Fron10der (changes requested):
1. I don't see the DEMO LINK
image

PR #3166, opened 2023-05-23
add task solution

h1deeeee:
- [DEMO LINK](https://h1deeeee.github.io/layout_catalog/)
- [TEST REPORT LINK](https://h1deeeee.github.io/layout_catalog/report/html_report/)


slavanosatskyi (changes requested):
there should be always 4 elements per row

Screenshot 2023-05-23 at 18 08 03

vlad-demchuk (changes requested):

nazarmatsevych (changes requested):
image

Try to fix your tests using margin instead of gap

nazarmatsevych (approved):

PR #3198, opened 2023-05-28
add task solution

purpleFade:
- [DEMO LINK](https://purpleFade.github.io/layout_catalog/)
- [TEST REPORT LINK](https://purpleFade.github.io/layout_catalog/report/html_report/)


SanyaBratashchuk (changes requested):
image

PR #3205, opened 2023-05-29
add task solution

adv555:
- [DEMO LINK](https://adv555.github.io/layout_catalog/)
- [TEST REPORT LINK](https://adv555.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,11 +2,384 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0"
+    >
     <title>Catalog</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    >
+    <link
+      rel="stylesheet"
+      href="./styles/main.scss"
+    >
   </head>
+
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__container">
+        <div class="header__content">
+          <div class="logo">
+            <a
+              href="#home"
+              class="logo__link"
+            >
+              <img
+                src="/images/logo.png"
+                alt="Moyo logo"
+                class="logo__image"
+              >
+            </a>
+          </div>
+          <nav class="nav">
+            <ul class="nav__list">
+              <li class="nav__item">
+                <a
+                  href="#apple"
+                  class="nav__link is-active"
+                >
+                  Apple
+                </a>
+              </li>
+              <li class="nav__item">
+                <a
+                  href="#samsung"
+                  class="nav__link"
+                >
+                  Samsung
+                </a>
+              </li>
+              <li class="nav__item">
+                <a
+                  href="#smartphones"
+                  class="nav__link"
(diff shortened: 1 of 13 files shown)


src/index.html
```diff
@@ -2,11 +2,447 @@
...
+
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="card__description">
+          Product code: 195434
+        </p>
+
+        <p class="card__rating">
```
juljaVysotska, line 141: I understand why you do that, but it's not a good idea to do that. Look, `p` - is a tag which used for text paragraphs. You used this tag not for text, you use it for an appearance that gives you p-tag. It's not text data, it's rating block with images. So, will be better if you use here `div` tags. You will write more styling, but on the other hand, structure will be more logical

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
...
+  padding: 32px 16px 16px;
+  width: $card-width;
+
+  border: 1px solid $border-color;
+  border-radius: 5px;
+
+  &__image {
+    display: block;
+    width: 160px;
+    height: 134px;
```
juljaVysotska, line 11: There will be good idea to add [objec-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) property for image. This property control how image will be scaled in size of your image will be different from size of your container

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  &__rating {
+    margin-bottom: 24px;
+
+    display: flex;
+    justify-content: space-between;
+    align-items: flex-end;
+
+    &-review {
```
juljaVysotska, line 40: `rating-review` is an element of `card` block. All elements must wrote inside a parent block element. Inside elements, we write pseudo-classes or pseudo-elements.

src/styles/main.scss
```diff
@@ -1,3 +1,13 @@
...
+@import "utils/mixins";
+
+@import "./helpers/container";
+
+@import "./base/global";
+
+@import "./blocks/header";
+@import "./blocks/logo";
+@import "./blocks/navigation";
+@import "blocks/catalog";
```
juljaVysotska, line 11: I think, you missed here `./` before `blocks/catalog`

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them. 

Best regards🌸

src/index.html
adv555: Hi! 👋 It was previously so. But in morning classes [Misha recommended ](https://mateacademy.slack.com/files/U054NT2QQKE/F05A291EZ63/image.png)changing div in that case for semantic tag p, so I refactored) Do I need to change it back as it was?

src/styles/blocks/card.scss
adv555: Added object-fit: cover ✅

src/styles/blocks/card.scss
adv555: Removed span as a redundant tag. Moved typography styles to card__rating ✅

src/styles/main.scss
adv555: Thx 😊 Fixed ✅

tetiana-ivanchenko (approved):

meforoma:
@adv555 header disappears on lower widths:
Catalog

cc @tetiana-ivanchenko

adv555:
👋 min-width for navigation is 1024, [but the catalog should use a grid for small screens.](https://github.com/mate-academy/layout_catalog/blob/master/grid.md)
Just for a better look added a burger menu icon for screens less than 1024px.

PR #3214, opened 2023-06-01
add task solution

GGsmou:
- [DEMO LINK](https://GGsmou.github.io/layout_catalog/)
- [TEST REPORT LINK](https://GGsmou.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,22 @@
+@import "./variables";
+
+@import "./mixins/transition";
+
+@import "./blocks/header";
+@import "./blocks/main";
+@import "./blocks/nav";
+@import "./blocks/stars.scss";
+@import "./blocks/card.scss";
+
+html {
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
+  font-size: 10px;
+  color: $main-text-color;
+}
+
 body {
   margin: 0;
+
+  background: $main-color;
+  overflow-x: hidden;
 }
--- a/src/styles/mixins/transition.scss
+++ b/src/styles/mixins/transition.scss
@@ -1,0 +1,3 @@
+@mixin transition($property, $duration: $transition) {
+  transition: $property $duration;
+}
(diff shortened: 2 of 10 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,22 @@
+@import "./variables";
+
+@import "./mixins/transition";
+
+@import "./blocks/header";
+@import "./blocks/main";
+@import "./blocks/nav";
+@import "./blocks/stars.scss";
+@import "./blocks/card.scss";
```
meforoma, line 9: some imports have no extension - some does.
Let's be consistent.)

src/styles/mixins/transition.scss
```diff
@@ -0,0 +1,3 @@
+@mixin transition($property, $duration: $transition) {
+  transition: $property $duration;
+}
```
meforoma, line 3: it could just be `mixins.scss`
and contain multiple mixins - if you had 5+ mixing you wouldn't create a file for each right?

src/styles/variables.scss
```diff
@@ -0,0 +1,15 @@
...
+
+$main-color: #fff;
+$header-color: #fff;
+$card-main-color: #fff;
+$card-secondary-color:#f3f3f3;
+$button-color: #00acdc;
+
+$main-text-color: #060b35;
+$text-secondary-color: #616070;
+$highlighted-text-color: #00acdc;
```
meforoma, line 11: > change the card title text color to #34568b when the card is hovered (.card:hover .card__title)

can't find `#34568b` around 🔍

meforoma (changes requested):
nice start.
checkout few comments in the code.

src/styles/variables.scss
```diff
@@ -0,0 +1,12 @@
+$main-color: #fff;
+$card-secondary-color:#f3f3f3;
+$button-color: #00acdc;
+
+$main-text-color: #060b35;
+$text-secondary-color: #616070;
+$highlighted-text-color: #00acdc;
```
tetiana-ivanchenko, line 7: no need to create a few variables with the same values

tetiana-ivanchenko (approved):

PR #3230, opened 2023-06-02
task layout catalog

Nyaroslav:
[DEMO LINK] - https://Nyaroslav.github.io/layout_catalog/
[TEST REPORT LINK] - https://Nyaroslav.github.io/layout_catalog/report/html_report/


src/index.html
```diff
@@ -4,9 +4,381 @@
...
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" 
+      rel="stylesheet"
+      crossorigin="anonymous"
+    >
   </head>
+
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <img
```
anastasiiavorobiova, line 29: There is a common HTML structure for the logo block:

```
<div class="header__logo">
  <a href="..." class="...">
    <img 
      src="..."
      alt="..."
      class="..."
     >
  </a>
</div>
```

The link should direct to the homepage of your website.
We also can add width and height attributes for the logo image in the header to avoid content flickering after CSSOM is built.
To know more about this you should read [the article about the critical rendering path and how the browser parses HTML files.](https://developer.mozilla.org/en-US/docs/Web/Performance/Critical_rendering_path)

src/index.html
```diff
@@ -4,9 +4,381 @@
...
+              video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="container">
+      <div class="container__holder">
+        <div class="card card--1" data-qa="card">
```
anastasiiavorobiova, line 107: Consider using more semantic tags for a card such as ```<article></article>```. It's semantic and fits better.

src/styles/main.scss
```diff
@@ -1,3 +1,214 @@
...
+    bottom: -0;
+    border-radius: 8px;
+    margin: 0 auto;
+    left: 0;
+    right: 0;
+    background: $header-text-color;
+  }
+}
+
+.container {
```
anastasiiavorobiova, line 200: don't meet task requirements:
1) there should always be 4 cards in a row (not 2, 3 or 5)
* use :nth-child(4n) to select every 4th element
* use :nth-last-child(-n + 4) to select 4 last elements
2) cards container should have fixed paddings bottom the same as top

anastasiiavorobiova (commented):
Nice work! Clear and readable code.
See my comments below

Volodymyr-Melnyk-Fron10der (changes requested):
deploy your solution

OleziO (approved):

PR #3284, opened 2023-06-06
add task solution

kylyvnyk:
DEMO LINK - https://kylyvnyk.github.io/layout_catalog/
TEST REPORT LINK - https://kylyvnyk.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,105 @@
+.card {
+  box-sizing: border-box;
+  width: 200px;
+  padding: 32px 16px 16px;
+  background: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  &__photo {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin: 0 auto;
+    margin-bottom: 40px;
+  }
+
+  &__title {
+    margin: 0;
+    margin-bottom: 4px;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+  }
+
+  &__code {
+    margin: 0;
+    margin-bottom: 16px;
+    color: $code-color;
+  }
+
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin: 0;
+    margin-bottom: 24px;
+  }
+
+  &__price {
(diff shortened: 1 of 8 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
...
+    text-transform: uppercase;
+    text-decoration: none;
+    border: 1px solid $main-color;
+    border-radius: 5px;
+    background-color: $main-color;
+
+    &:hover {
+      color: $main-color;
+      background-color: #fff;
+      transform: scale(1.2);
```
tetiana-ivanchenko, line 78: 1. hover effects should have transition
2. according to mockup bitton shouldn't scale on hover

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,105 @@
...
+    height: 16px;
+    width: 16px;
+    background-image: url(../images/star.svg);
+    background-position: center;
+    background-repeat: no-repeat;
+
+    &:not(:last-child) {
+      margin-right: 4px;
+
+    }
```
tetiana-ivanchenko, line 97:
```suggestion
      margin-right: 4px;
    }
```

tetiana-ivanchenko (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,111 @@
...
+    transition: all 300ms ease-in-out;
+
+    &:hover {
+      color: $main-color;
+      background-color: #fff;
+    }
+  }
+}
+
+.stars {
```
anstsot, line 90: it would be better to move `stars` BEM block to another file (one block = one file)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,9 @@
+.catalog__content {
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: repeat(2, 406px);
```
anstsot, line 4: don't hardcode the height of cards. this property can be removed so the catalog can have unlimited number of rows

anstsot (approved):
Good work, approving:sparkles:
But pay attention to my comments to avoid the same mistakes in the future

PR #3311, opened 2023-06-15
add task solution

vpMariia:
https://vpMariia.github.io/layout_catalog/
https://vpMariia.github.io/layout_catalog/report/html_report/

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,0 +1,60 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  min-width: 1024px;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,60 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  min-width: 1024px;
```
etojeDenys, line 6: looks strange, why do you need to use here `min-width`?

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #3346, opened 2023-06-30
add task solution for catalog

PavloMykhalov:


volodymyr-soltys97 (changes requested):
Try to fix problems with test. You need to add `margin-top` for the tag main. 
But add the class first and do it via the class selector
image

src/styles/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    height: 16px;
+    margin-top: 16px;
+
+  }
+
+  &__reviews {
+    font-weight: 400;
```
sTorba24, line 48: can be removed as default value

src/styles/navigation.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.nav__link {
+  text-decoration: none;
+  text-transform: uppercase;
+  color: $nav-link-color;
+  font-weight: 500px;
+  font-size: 12px;
+  line-height: 14px;
+  height: 60px;
```
sTorba24, line 19: don't set height directly, use paddings instead

src/styles/stars.scss
```diff
@@ -0,0 +1,21 @@
+.stars {
+  display: flex;
+
+  &__star {
+    background-image: url(../images/star.svg);
+    background-repeat: no-repeat;
+    background-position: center;
```
sTorba24, line 7: use background shorthand property

sTorba24 (changes requested):
Almost good! Check font size on header to pass a test

anastasiiavorobiova (approved):
Awesome &#128293;

PR #3364, opened 2023-07-07
add incomplete solution

LabPetro:
[DEMO LINK](https://LabPetro.github.io/layout_catalog/)

[TEST REPORT LINK](https://LabPetro.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Frontend practice with catalog page
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_catalog/report/html_report/)
+- [DEMO LINK](https://LabPetro.github.io/layout_catalog/)
+- [TEST REPORT LINK](https://LabPetro.github.io/layout_catalog/report/html_report/)
 
 > Follow [this instruction](https://github.com/mate-academy/layout_task-guideline#how-to-solve-the-layout-tasks-on-github)
 ___


Anastasiia-Svintsova (changes requested):
- Continue conversation in the chat

src/styles/main.scss
```diff
@@ -1,3 +1,16 @@
+@import "./utils/_variables";
+@import "./utils/_mixins";
+@import "./blocks/container.scss";
+@import "./blocks/header.scss";
+@import "./blocks/nav.scss";
+@import "./blocks/card.scss";
+@import "./blocks/stars.scss";
+
+* {
```
lerastarynets, line 9: Pls make separate `reset.scss` file for `*` and `body` styles

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  @include dimensions;
+  background-image: url(../images/star.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+  margin-right: 4px;
```
lerastarynets, line 10: The last child of '.stars__star' will also have a 'margin-right'. You might want to remove it to keep the content aligned properly. You can use ':not(:last-child)' to select all elements except the last one.

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,46 @@
...
+    text-transform: uppercase;
+    text-decoration: none;
+  }
+
+  &__link:hover,
+  .is-active {
+    color: $color-blue;
+  }
+
+  .is-active::before {
```
lerastarynets, line 36: Similarly, the `.is-active::before` class should also be nested within `&__link`. Otherwise, it won't be applied to the correct element.

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,46 @@
...
+    font-weight: $medium-font-weight;
+    font-size: $medium-font-size;
+    color: $color-main-dark;
+    text-transform: uppercase;
+    text-decoration: none;
+  }
+
+  &__link:hover,
+  .is-active {
+    color: $color-blue;
```
lerastarynets, line 33: The `.is-active` class should be nested within `&__link`, as it's used to indicate the active link. Currently, the hover effect will not apply to the active link.

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin dimensions {
```
lerastarynets, line 1: The name of the mixin 'dimensions' is too generic. It would be better to give it a more specific name relating to its purpose, for example 'starDimensions'.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,21 @@
...
+  justify-content: space-between;
+  position: relative;
+  height: 60px;
+  width: 100%;
+  background-color: $color-bright;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05) ;
+
+  &__link {
+    display: flex;
+    margin: 10px 0 10px 50px;
```
lerastarynets, line 19: The margin property on the '.header__link' selector could potentially lead to inconsistent layout. If you want to center the element, consider using flexbox properties to center it vertically within the '.header' element instead of using margins.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,21 @@
+.header,
+.nav__list {
```
lerastarynets, line 2: The '.nav__list' styles are declared in the 'header.scss' file. It's a good practice to keep the styles for different blocks in separate files. Consider moving '.nav__list' styles to 'nav.scss'.

src/index.html
```diff
@@ -5,8 +5,439 @@
...
+    <main class="container">
+      <div class="card container__card" data-qa="card">
+        <img
+          class="card__image"
+          src="images/imac.jpeg"
+          alt="APPLE A1419 iMac 27'' Retina 5K Monoblock (MNED2UA/A)"
+        >
+
+        <p class="card__product-name">
+           APPLE A1419 iMac 27" Retina <br>
```
lerastarynets, line 112: Avoid using` <br>` to force line breaks or spacing. That's a job for CSS. If it's a paragraph, then make it a paragraph `<p>` or `<div>`

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,24 @@
...
+
+  & > :nth-child(4n) {
+    margin-right: $reset;
+  }
+
+  & > :nth-last-child(-n+4) {
+    margin-top: 46px;
+  }
+
+  @media screen and (min-width: 1024px) and (max-width: 1200px) {
```
lerastarynets, line 21: In the media query, the width is being set to 1024px. This may cause horizontal scrolling on devices with a viewport width of less than 1024px. Consider using a grid

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,24 @@
...
+  &__card {
+    margin-right: 48px;
+  }
+
+  & > :nth-child(4n) {
+    margin-right: $reset;
+  }
+
+  & > :nth-last-child(-n+4) {
+    margin-top: 46px;
```
lerastarynets, line 18: The selector :nth-last-child(-n+4) might not behave as expected as it will apply the styles to the last 4 elements, regardless of their type. Make sure that you have less than 4 different types of elements inside .container or specify the element type.

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,24 @@
+.container {
+  margin: 0 auto;
+  width: $container-width;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  padding: 50px 0;
+
+  &__card {
+    margin-right: 48px;
```
lerastarynets, line 10: The margin-right property may not push the fourth card to the next line if the width of the container is not enough. It's recommended to use gap property with grid instead of margin-right.

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
+.stars {
+  display: flex;
+}
+
+.stars__star {
```
lerastarynets, line 5: pls make use of sass nesting

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
...
+  background-repeat: no-repeat;
+  background-position: 50%;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:first-child,
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
```
lerastarynets, line 17: better ti use `for` loop here

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,17 @@
...
+$medium-font-size: 12px;
+$button-font-size: 14px;
+$large-font-size: 16px;
+$color-blue: #00acdc;
+$color-main-dark: #060b35;
+$color-secondary-grey: #616070;
+$color-bright: #fff;
+$gap-small: 16px;
+$gap-large: 24px;
+$reset: 0;
```
lerastarynets, line 17: the purpose of `reset` variable is not clear because of its generic name

src/styles/utils/_variables.scss
lerastarynets, line 1: Pls reduce the number of variables, you have too many of them, no need to create variables if you use them only once in your project

src/index.html
```diff
@@ -5,8 +5,439 @@
...
+        </p>
+
+        <p
+          class="card__product-code"
+        >
+           Product code: 195434
+        </p>
+        <div class="card__stars-wrapper">
+          <div class="stars stars--4 card__stars-block">
+            <div class="stars__star"></div>
```
lerastarynets, line 292: pls blank lines between children elements
```suggestion
            <div class="stars__star"></div>
            
```

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist
Also pls make use of `transitions` to make all the hover effects on the page smooth

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,102 @@
...
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 0 16px;
+  color: $color-main-dark;
+  transition-property: transform;
+  transition-duration: 0.3s;
+
+  &:hover {
+    transition-duration: 0.3s;
+    transform: scale(1.2);
```
polosanya, line 13:
```suggestion
  &:hover {
    transform: scale(1.2);
```
There is no need to add transition for hover state

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  &__product-name {
+    margin-top: $margin-zero;
+    margin-bottom: 4px;
+    font-size: $medium-font-size;
+    font-weight: $medium-font-weight;
+    line-height: 18px;
+
+    &:hover {
```
polosanya, line 31: Let's add transition here too

polosanya (changes requested):
Almost perfect! Make sure that all hover effects have transition.
And would be great to fix the moments discussed on Q&A (function, margins etc)

lerastarynets (approved):
All requirements are met. You're on the right track!

PR #3381, opened 2023-07-11
Develop

domalewski:
[DEMO LINK](https://domalewski.github.io/layout_catalog/)
[TEST REPORT LINK](https://domalewski.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,146 @@
+@import "../utils/variables.scss";
+
+.catalog {
+  padding-top: 50px;
+  display: grid;
+  grid-template-columns: repeat($column-count, $width-card);
+  grid-template-rows: repeat(2, 1fr);
+  gap: 46px 48px;
+  justify-content: center;
+
+  @media (max-width: 487px) {
+    $column-count: 1;
+    grid-template-columns: repeat($column-count, $width-card);
+  }
+
+  @media (min-width: 488px) {
+    $column-count: 2;
+    grid-template-columns: repeat($column-count, $width-card);
+  }
+
+  @media (min-width: 768px) {
+    $column-count: 3;
+    grid-template-columns: repeat($column-count, $width-card);
+  }
+
+  @media (min-width: 1024px) {
+    $column-count: 4;
+    grid-template-columns: repeat($column-count, $width-card);
+  }
+}
+
+.card {
+  width: $width-card;
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  transition: 0.3s;
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
+  &__image-container {
+    margin: 0 3px 38px;
+  }
+
+  &__image {
+    width: 100%;
+    height: 98%;
+    object-fit: contain;
+  }
+
+  // &__content {
+  //   padding: 4px 16px 16px;
+  // }
+
+  &__title {
+    margin-bottom: 4px;
+    font-family: $font-family;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $color-main-accent;
+  }
+
+  &__product-code {
+    margin-bottom: 16px;
+    font-family: $font-family;
+    font-weight: 400;
+    font-size: 10px;
+    color: $color-secondary;
+  }
+
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    font-family: $font-family;
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    color: $color-main-accent;
+  }
+
+  // &__review {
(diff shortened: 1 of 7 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,8 @@
+$color-main-accent: #060b35;
+$color-secondary: #616070;
+$color-accent-blue: #00acdc;
+$color-white: #fff;
+$color-hover-text: #34568b;
+$font-family: roboto, helvetica, sans-serif;
+$column-count: 4;
```
mvjl000, line 7: I think this variable is unnecessary

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,142 @@
...
+  display: grid;
+  grid-template-columns: repeat($column-count, $width-card);
+  grid-template-rows: repeat(2, 1fr);
+  gap: 46px 48px;
+  justify-content: center;
+
+  @media (max-width: 487px) {
+    $column-count: 1;
+    grid-template-columns: repeat($column-count, $width-card);
+  }
```
mvjl000, line 14: No need for this media query. Leave default property in catalog block as it is and the next media query will overwrite it

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,142 @@
...
+
+  @media (min-width: 488px) {
+    $column-count: 2;
+    grid-template-columns: repeat($column-count, $width-card);
```
mvjl000, line 18: It would be better if you didn't use the variable here and simply do
```suggestion
    grid-template-columns: repeat(2, $width-card);
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,142 @@
...
+
+.card {
+  width: $width-card;
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  transition: 0.3s;
```
mvjl000, line 40: List all properties that you are applying transition to (don't use `all` either)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,142 @@
...
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
+  &__image-container {
+    margin: 0 3px 38px;
```
mvjl000, line 47: Maybe instead of 3px on sides use `auto`?

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  &__image {
+    width: 100%;
+    height: 98%;
+    object-fit: contain;
```
mvjl000, line 53: Try to set image dimensions same as in figma. Maybe it will fix the image diff issue in the test report

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,8 @@
+$color-main-accent: #060b35;
+$color-secondary: #616070;
+$color-accent-blue: #00acdc;
+$color-white: #fff;
+$color-hover-text: #34568b;
+$font-family: roboto, helvetica, sans-serif;
```
mvjl000, line 6: You are setting the font-family in every possible place. Simply put it in body

mvjl000: Same thing with other variables. If value of a property is inherited and you want it to be so, don't apply this value again

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  &__star {
+    height: 16px;
+    width: 16px;
+    background-repeat: no-repeat;
+    background-position: center;
+    background-image: url(../images/star.svg);
+    margin-right: 4px;
+
+    :last-child {
```
mvjl000, line 12: I think you are missing the `&` here
```suggestion
    &:last-child {
```

mvjl000 (changes requested):
In addition to the fixes from my comments, try to figure out why there are text differences throughout the test report on the entire page

domalewski:
the font is still incorrect. I tired different approaches to fix this. So maybe this is the problem that my whole layout is all over the place

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,10 @@
...
```
mvjl000, line 6: There is something wrong with your default font. In your deployed version you are not using the Roboto. Maybe that's why some of your tests fail. Try to change it ;)

Conclusion: Make roboto a default font

mvjl000 (changes requested):

choeqq (changes requested):
So now we have only 3 tests left, I think if you make your fonts bolder/thinner we're good to go! 🚀

choeqq (changes requested):
Looks good, but please try to fix `font-weight` on `Laptops & Computers` nav link and you're good go 🚀

JoyCoffeeAddict (approved):
Approve, since it's rather a problem with tests than with a page itself

PR #3383, opened 2023-07-11
add task solution

marekder:


src/index.html
```diff
@@ -5,8 +5,291 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#home">
+          <img src="images/logo.png" alt="logo">
```
mvjl000, line 16: Remember about using more descriptive alt attributes (for SEO). "Moyo logo" would be much better ;)

src/index.html
```diff
@@ -5,8 +5,291 @@
...
+          <p class="price__label">Price:</p>
+          <p class="price__value">$2,199</p>
+        </div>
+        <a
+          href="#"
+          data-qa="hover"
+          class="card__button"
+          >Buy </a>
+      </div>
+      <div class="card" data-qa="card">
```
mvjl000, line 85: This data-qa is supposed to be only in the first card (Check the readme)

src/index.html
```diff
@@ -5,8 +5,291 @@
...
+          <div class="card__rate-reviews">Reviews: 5</div>
+        </div>
+        <div class="price">
+          <p class="price__label">Price:</p>
+
+          <p class="price__value">$2,199</p>
+        </div>
+        <a
+          href="#"
+          data-qa="hover"
```
mvjl000, line 110: Same thing here, only first card should have this data-qa

mvjl000: Make sure to remove it from the other cards ;)

src/styles/main.scss
```diff
@@ -1,3 +1,23 @@
...
+  margin: 0;
+}
+
 body {
+  font-family: Roboto, Helvetica, sans-serif;
+}
+
+p {
   margin: 0;
 }
```
mvjl000, line 23: You are already setting margin 0 in line 10

src/blocks/catalog.scss
```diff
@@ -0,0 +1,23 @@
+.catalog {
+  display: grid;
+  justify-content: center;
+  padding: 50px 0;
+  column-gap: 48px;
+  row-gap: 46px;
+
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, auto);
+  }
```
mvjl000, line 10: This property doesn't need a media query. When the viewport hits 488px, the next media query's property will overwrite this one.

mvjl000 (changes requested):
I have left a few comments but I guess that your is going to need a couple more adjustments later on anyway.

For now very important thing, change the structure of your styles:

- move blocks folder to styles
- reduce the number of blocks - you don't need so many of them. All you need is block for header, nav, catalog, card. (For example - instead of creating price block just do `card__price-container` and inside `card__price-label` and `card__price`)

src/index.html
mvjl000: @marekder one thing I forgot about - add smooth transitions and animations. Check the readme

src/blocks/catalog.scss
marekder: Right, removed.

src/index.html
marekder: Done.

src/index.html
marekder: Done.

src/index.html
marekder: Done.

src/styles/main.scss
marekder: Removed.

src/index.html
```diff
@@ -5,8 +5,288 @@
...
+          <div class="card__rate-reviews">Reviews: 5</div>
+        </div>
+        <div class="card__price-container">
+          <p class="card__price-label">Price:</p>
+
+          <p class="card__price">$2,199</p>
+        </div>
+        <a
+          href="#"
+          data-qa="hover"
```
asakevych, line 139: this data-qa should be only for 1st card, please remove it everywere else

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,117 @@
...
+      font-size: 12px;
+      color: $details-color;
+
+    }
+
+    &-value {
+      font-weight: 700;
+      font-size: 16px;
+      color: $font-color;
+    }
```
asakevych, line 64: those styles doesn't work 
using `&` you mean that you take name of parent and add some ending, so here you have `card__price-container-label` and you doesn't have element with such class
it should be: 
```suggestion
&__price {

    &-container {
      ...
    }

    &-label {
     ...
    }

    &-value {
      ...
    }
}
```

asakevych (changes requested):
Good job, but tests are not passed, try to make them pass and request reiew once more 😉 
1. Please add links to demo and tests
2. check if you have everything inside card he same as is in the mock
3. add transition to button colors change
4. check once more all fonts, in tests it seems you have something wrong with them
5. please check checklist and requires for data-qa once more

src/index.html
marekder: Done.

src/styles/blocks/card.scss
marekder: Corrected.

choeqq (changes requested):
Test aren't passing still, it looks like your font is little too bold 🤔 

Code looks really good, try to fix font problems and you're good to go!

marekder:
> Test aren't passing still, it looks like your font is little too bold 🤔
> 
> Code looks really good, try to fix font problems and you're good to go!

Unfortunately I am not able to find reason for different fonts, got no clue what's going on

PR #3410, opened 2023-07-14
add header with catalog of goods

FrontKid:
- [DEMO LINK](https://FrontKid.github.io/layout_catalog/)
- [TEST REPORT LINK](https://FrontKid.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/1-helpers/_common.scss
+++ b/src/styles/1-helpers/_common.scss
@@ -1,0 +1,15 @@
+html {
+  font-size: $fz;
+  font-family: "Roboto", "Helvetica", sans-serif;
+  line-height: $lh;
+  color: $black;
+  font-weight: 400;
+  font-style: normal;
+}
+
+.page__container {
+  max-width: 1200px;
+  padding-left: 50px;
+  padding-right: 50px;
+  margin: auto;
+}
--- a/src/styles/1-helpers/_common.scss
+++ b/src/styles/1-helpers/_common.scss
@@ -7,6 +7,7 @@ html {
   font-style: normal;
 }
 
+.header__container,
 .page__container {
   max-width: 1200px;
   padding-left: 50px;
--- a/src/styles/1-helpers/_mixins.scss
+++ b/src/styles/1-helpers/_mixins.scss
@@ -1,0 +1,5 @@
+@mixin flex-centering($just-cont) {
+  display: flex;
+  justify-content: $just-cont;
+  align-items: center;
+}
--- a/src/styles/1-helpers/_variables.scss
+++ b/src/styles/1-helpers/_variables.scss
@@ -1,7 +1,7 @@
 // colors
 $white: #fff;
 $blue: #00acdc;
-$black: #000;
+$black: #060b35;
 $grey: #616070;
 
 // general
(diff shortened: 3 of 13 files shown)


src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
+// colors
+$white: #fff;
+$blue: #00acdc;
+$black: #060b35;
+$grey: #616070;
```
roman-mirzoian, line 5: When naming variables, you should use the name according to the purpose, and not according to the color, because the color may change to another during the development process. Then you have to update it throughout the code, not just where this variable is created.

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
...
+
+// general
+$fz: 12px;
+$lh: 18px;
```
roman-mirzoian, line 9: Variable names should be given in such a way that it is clear what they are for in any context.
Therefore, it is not necessary to shorten the names to unreadable ones.

src/styles/1-helpers/_common.scss
```diff
@@ -0,0 +1,16 @@
+html {
+  font-size: $fz;
+  font-family: "Roboto", "Helvetica", sans-serif;
```
roman-mirzoian, line 3: Now everything is done correctly, but it is a good practice to also put the `font-family` in a separate variable.

src/styles/1-helpers/_mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin flex-centering($just-cont) {
```
roman-mirzoian, line 1: Your `flex-centering` mixin doesn't do centering, so it would be better to call it `flex-align` (i.e. a more generic name).

roman-mirzoian (changes requested):
The work is done well, but there is still a little improvement in the code.

src/styles/3-elements/_navbar.scss
```diff
@@ -0,0 +1,47 @@
+.nav {
+  min-width: 635.54px;
```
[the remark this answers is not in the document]
FrontKid: if I remove max-width and give items "margins" is It not the same thing like if I give the nav container min-width and 100% for list?

src/styles/2-components/_header.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.logo__img {
+  display: block;
+}
+
+.header-container {
+  @include flex-align(space-between);
+  box-sizing: border-box;
+  font-weight: 500;
+}
```
BudnikOleksii, line 13: If it is a separate block "logo", move to the file logo, if it's related to the header the class name should be .header__logo-img. The same for .header-container

src/styles/3-UI/_card-button.scss
```diff
@@ -0,0 +1,24 @@
+.card-button {
+  display: block;
+  border-radius: 5px;
+  background-color: $button-background;
+  padding-top: 10.2px;
+  padding-bottom: 10.2px;
```
BudnikOleksii, line 6: Avoid float numbers for pixels. Also, you can use shorthand property padding instead.

BudnikOleksii (approved):
Nicely done🔥
Just remember to create a separate file for each BEM block.

PR #3458, opened 2023-07-24
add task solution

AndrMar1939:
- [DEMO LINK](https://AndrMar1939.github.io/layout_catalog/)
- [TEST REPORT LINK](https://AndrMar1939.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,12 +1,476 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet">
   </head>
-  <body>
-    <h1>Catalog</h1>
+  <body class="page__body">
+    <header class="header page__header">
+      <a href="#home" class="header__link">
+        <img
+          src="images/logo.png"
+          alt="logo"
+          class="header__logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">
+              Samsung
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">
+              Smartphones
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#laptops&computers"
+              class="nav__link"
+              data-qa="nav-hover">
+              Laptops & Computers
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">
+              Gadgets
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#tablets" class="nav__link">
+              Tablets
+            </a>
+          </li>
+          <li class="nav__item">
+            <a href="#photo" class="nav__link">
+              Photo
+            </a>
+          </li>
+          <li class="nav__item">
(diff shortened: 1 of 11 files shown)


src/index.html
```diff
@@ -1,12 +1,476 @@
...
+          class="header__logo">
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
```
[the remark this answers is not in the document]
Moroz-Dmytro: Fix in all other places
image

src/styles/blocks/nav-list.scss
```diff
@@ -0,0 +1,34 @@
...
+    list-style: none;
+  }
+
+  &__item {
+    margin-right: 20px;
+  }
+
+  &__item:last-child {
+    margin-right: 0;
+  }
```
Moroz-Dmytro, line 15: Nesting

src/styles/blocks/nav-list.scss
```diff
@@ -0,0 +1,34 @@
...
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 14px;
+    text-decoration: none;
+    text-transform: uppercase;
+  }
+
+  &__link:hover {
+    color: $button-color;
+  }
```
Moroz-Dmytro, line 33: Nesting

Moroz-Dmytro (changes requested):
Good progress 👍 
But what about

image


image

src/styles/main.scss
```diff
@@ -1,3 +1,27 @@
...
+.is-active::after {
+  content: "";
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  background-color: $button-color;
+  height: 4px;
+  border-radius: 8px;
 }
```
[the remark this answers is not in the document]
AndrMar1939: omg

src/styles/blocks/nav-list.scss
```diff
@@ -0,0 +1,39 @@
+.nav {
+  // hide navbar for properly view for catalog with grid
+  @media (max-width: 800px) {
+    display: none;
+  }
```
alexander-ignatow, line 5: Good try, but on smaller screens, you are removing navigation completely - that is not perfect
You have either to add some navigation for smaller screens (that is out of the scope here) or just think of something else

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+    color: $secondary-font-color;
+  }
+
+  &__price-value {
+    @include zero-margin;
+    @include large-font-size;
+  }
+
+  &__link {
+    width: 166px;
```
alexander-ignatow, line 69: avoid hardcoding width - unless you need it
think of some other option - just to let button fill the parent

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
+.card {
+  @include flex-column-centred;
+
+  max-width: $card-width;
```
alexander-ignatow, line 4: I would suggest changing it to max-width

Per task, you need 200px cards, while max-width does not guarantee the width
so `width:$card-width` look like cleaner solution here

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  @media (min-width: 768px) {
+    --cells: 3;
+  }
+
+  @media (min-width: 1024px) {
+    --cells: 4;
+  }
+}
```
alexander-ignatow, line 22: Interesting idea implementing it with variable 👍

src/styles/main.scss
```diff
@@ -1,3 +1,24 @@
...
+    content: "";
+    position: absolute;
+    bottom: 0;
+    right: 0;
+    left: 0;
+    background-color: $button-color;
+    height: 4px;
+    border-radius: 8px;
+  }
 }
```
alexander-ignatow, line 24: I don't see why you are setting this class here
It should be moved to the corresponding file inside the block

If you are reusing this - move it to mixin

alexander-ignatow (changes requested):
Overall, a good solution 👍 

Just fix some minor issues to make it perfect!

Moroz-Dmytro (approved):
LGTM 👍

PR #3462, opened 2023-07-24
add task solution

polinavafik:
- [DEMO LINK](https://polinavafik.github.io/layout_catalog/)
- [TEST REPORT LINK](https://polinavafik.github.io/layout_catalog/report/html_report/)

**question**
where would it be better to place .is-active styles? in a separate .scss file?
because I tried to rename it to .nav__link--is-active so that it would be inside the nav styles file, but then the tests wouldn't work :)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,8 @@
+$color-main-accent: #060b35;
+$color-secondary: #616070;
+$color-accent-blue: #00acdc;
+$color-hover-button: white;
+$color-hover-title: #34568b;
+$navElementHeight: 23px;
+$fontFamily:"Roboto", "Helvetica", sans-serif;
+$transitionTime: 0.3s;
```
roman-mirzoian, line 8: Try to maintain a single naming style for variables ( only `kebab-case` or only `camelCase`).
It is also better to place the variable type at the end ( `$main-accent-color`, `$secondary-color` ).

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,53 @@
...
+    white-space: nowrap;
+
+    &:hover {
+      transition: $transitionTime;
+      color: $color-accent-blue;
+    }
+  }
+}
+
+.is-active {
```
roman-mirzoian, line 40: Regarding your question about placing the `.is-active` class: it is better not to put (nested) it in other classes, but to keep it in the same file as the class for which you want to use it.
So you did everything right 👍

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/styles/utils/variables.scss
polinavafik: okay, i'll change everything to kebab-case

about the color naming, I just used the example from the checklist so that's why its like that :)
BAD example:
```scss
$main: #060b35;
$secondary: #616070;
$blue: #00acdc;
```
GOOD example:
```scss
$color-main-accent: #060b35;
$color-secondary: #616070;
$color-accent-blue: #00acdc;
```

Moroz-Dmytro (approved):
LGTM 🔥

PR #3474, opened 2023-07-24
Catalog

vatatan:
- [DEMO LINK](https://vatatan.github.io/layout_catalog/)
- [TEST REPORT LINK](https://vatatan.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -1,0 +1,101 @@
+.card {
+  padding: 16px;
+  width: 200px;
+  border-radius: 5px;
+  border: 1px solid $borderColor;
+  transition: all $transition-time;
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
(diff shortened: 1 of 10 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,101 @@
+.card {
+  padding: 16px;
+  width: 200px;
+  border-radius: 5px;
+  border: 1px solid $borderColor;
+  transition: all $transition-time;
```
Anna-Tkachenko, line 6: Don't use `all` for transition property, specify only the needed ones 
As `all` affects in the bad way on the website performance

Anna-Tkachenko (approved):
Clean code and great solution, well done! ⭐️

src/styles/card.scss
vatatan: done

PR #3481, opened 2023-07-25
add task solution

NazarVotkanych:


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+
+    text-transform: uppercase;
+    color: $mainColor;
+    font-weight: 700;
+    text-decoration: none;
+
+    &:hover {
+      color: $activeColor;
+      background-color: $buttonBackground;
+      transition: $transitionTime;
```
Anna-Tkachenko, line 82: Let's move transition property to the main class, not on the hover state
And don't forget to specify the property, for which you want to apply the transition

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,74 @@
...
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  flex-direction: row;
+
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+}
+
+.logo {
```
Anna-Tkachenko, line 11: Is it the separate BEM block? Looks like it should not be in the file for header styles

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,74 @@
...
+  margin-left: 50px;
+  margin-top: 5px;
+}
+
+.nav {
+  margin-right: 50px;
+}
+
+.nav__list {
+  display: flex;
```
Anna-Tkachenko, line 23: Let's create separate file for the styles for `nav` BEM block and don't forget about nesting

Anna-Tkachenko (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+      background-color: $buttonBackground;
+    }
+  }
+
+  &:hover {
+    transition: $transitionForAnimation;
+    transform: scale(120%);
+
+    .card__title {
+      transition: $transitionForAnimation;
```
BudnikOleksii, line 91: Apply transitions to the general selector so it will work both ways, not just from hover, check it in all places.

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,57 @@
+.nav {
+  margin-right: 50px;
```
BudnikOleksii, line 2: BEM block couldn't have any margins.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,16 @@
...
+  justify-content: space-between;
+  flex-direction: row;
+
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+
+  &__logo {
+    display: flex;
+    align-items: center;
+    margin-left: 50px;
+    margin-top: 5px;
```
BudnikOleksii, line 14: Let's use shorthand property margin instead. Also, it would be better to control these spaces with padding in the header element, in this case, you can get rid of the margin-right in navigation.

BudnikOleksii (changes requested):
Looks great🔥
Just get rid of margins in the BEM block. And fix the transitions in all places.

src/index.html
```diff
@@ -4,9 +4,435 @@
...
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#home">
+          <img src="./images/logo.png" alt="Logo">
```
Moroz-Dmytro, line 16:
```suggestion
          <img src="./images/logo.png" alt="Moyo-logo">
```

src/index.html
```diff
@@ -4,9 +4,435 @@
...
+        <img
+          src="./images/imac.jpeg"
+          alt="APPLE A1419 iMac 27 Retina 5K Monoblock (MNED2UA/A)"
+          class="card__image"
+        >
+
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <p class="card__code">
```
Moroz-Dmytro, line 73:
```suggestion
        </h2>
        
        <p class="card__code">
```

Moroz-Dmytro: And in other places

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    margin-bottom: 24px;
+  }
+
+  &__reviews-label {
+    display: flex;
+    align-items: center;
+    font-size: 10px;
+    line-height: 14px;
+    color: $textColor;
+  }
```
Moroz-Dmytro, line 44: Nesting

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    color: $textColor;
+    transition: $transitionForAnimation;
+  }
+
+  &__code {
+    color: $secondaryColor;
+    font-size: 10px;
+    line-height: 14px;
+    margin-bottom: 16px;
+
```
Moroz-Dmytro, line 29:
```suggestion
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,95 @@
...
+    line-height: 14px;
+    color: $textColor;
+  }
+
+  &__price-container {
+    display: flex;
+    justify-content: space-between;
+
+    margin-bottom: 16px;
+
```
Moroz-Dmytro, line 51:
```suggestion
```

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,15 @@
+
```
Moroz-Dmytro, line 1:
```suggestion
```

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,59 @@
...
+  &__link.is-active {
+    color: $activeColor;
+  }
+
+  &__link:hover {
+    color: $activeColor;
+    transition: 0.2ms;
+  }
+
+  &__link.is-active::after {
```
Moroz-Dmytro, line 44: Nesting

Moroz-Dmytro: Also don't increase specificity if no needed. Move `.is-active` class up from this block(use only `.is-active` without `&__link`)

Moroz-Dmytro (changes requested):
Some comments to make it perfect 😉

src/styles/blocks/grid_container.scss
```diff
@@ -0,0 +1,27 @@
+.container {
+  --column-count: 1;
```
roman-mirzoian, line 2: It is better to use SCSS variables (`$column-count`).

src/styles/blocks/grid_container.scss
```diff
@@ -0,0 +1,27 @@
...
+  .container {
+    --column-count: 3;
+  }
+}
+
+@media (min-width: 1024px) {
+  .container {
+    --column-count: 4;
+  }
+}
```
roman-mirzoian, line 27: Now everything works perfectly, but it is better not to change the value of the variable depending on the size, but to change the parameters themselves.

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,93 @@
+.card {
+  width: 200px;
+  padding: 32px 16px 16px;
+  border: 1px solid $mainColor;
+  border-radius: 5px;
+  transition: $transitionForAnimation;
```
roman-mirzoian, line 6: For `transition`, it is also worth specifying what exactly to use it for.
https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

roman-mirzoian (approved):
Well done, but pay attention to the last comments.

PR #3482, opened 2023-07-25
Develop

edgergard:
- [DEMO LINK](https://edgergard.github.io/layout_catalog/)
- [TEST REPORT LINK](https://edgergard.github.io/layout_catalog/report/html_report/)


src/styles/header.scss
alexander-ignatow, line 1: Can you add some space here ?

image

src/styles/card.scss
```diff
@@ -0,0 +1,131 @@
...
+    font-size: 14px;
+    font-weight: 700;
+    line-height: 16px;
+    color: #fff;
+    text-decoration: none;
+    background-color: $secondaryColor;
+    transition-duration: $animationTransitionDuration;
+
+    &:hover {
+      border: 1px solid $secondaryColor;
```
alexander-ignatow, line 97: You can set this border to button and don't change it on hover

src/styles/card.scss
```diff
@@ -0,0 +1,131 @@
...
+  &:hover {
+    transform: scale(1.2);
+
+    .card__top-title {
+      color: #34568b;
+    }
+  }
+}
+
+.stars {
```
alexander-ignatow, line 113: BEM blocks should be separate files

src/styles/catalog.scss
```diff
@@ -0,0 +1,8 @@
+.catalog {
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  row-gap: 46px;
+  column-gap: 48px;
+  padding-top: 50px;
+}
```
alexander-ignatow, line 8: Al long as you are using grid, see the task:

> Catalog using grid layout
> Implement the layout using grid
>
> Gap between cards should be: horizontal - 48px, vertical - 46px
> Blocks should start from the left
> Media queries:
> There should be four columns starting at 1024px
> There should be three columns starting at 768px
> There should be two columns starting at 488px
> There should be one column below 488px

alexander-ignatow: In addition - if it is ok with the task, better to use flexible column sizes (in `fr`)
If you need columns to be exactly `200px` consider using a variable (because it is the same as card width)

alexander-ignatow (changes requested):
Very good! 👍 

Please see the comments to improve your code + you have missed a minor thing on the grid layout

src/styles/header.scss
alexander-ignatow: 👍

src/styles/header.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  font-size: 12px;
+  font-weight: 500;
+  line-height: 100%;
+
+}
+
+.nav {
+  margin-left: 50px;
+
```
alexander-ignatow, line 15: Per BEM:

1 Each block should be in a separate file
2 No margins on BEM blocks

src/styles/header.scss
```diff
@@ -0,0 +1,59 @@
...
+    &:hover {
+      color: $secondaryColor;
+    }
+  }
+}
+
+.is-active {
+  color: $secondaryColor;
+
+  &::after {
```
alexander-ignatow, line 47: You are treating `is-active` as a block

If it is a block - it should be in a separate file
But why you have made it a block? Please explain, if you will keep it a block.

src/styles/stars.scss
```diff
@@ -0,0 +1,19 @@
+.stars {
+  display: flex;
+  margin-left: 4px;
```
alexander-ignatow, line 3: No margins for blocks

src/styles/stars.scss
```diff
@@ -0,0 +1,19 @@
...
+    height: 16px;
+    margin-right: 4px;
+    background: url(../images/star.svg);
+    background-repeat: no-repeat;
+  }
+
+  @for $i from 1 through 5 {
+    &--#{$i} &__star:nth-child(-n + #{$i}) {
+      background: url(../images/star-active.svg);
+      background-repeat: no-repeat;
```
alexander-ignatow, line 16: Do you really need to set `background-repeat` here?

alexander-ignatow (changes requested):
Good! You have fixed many issues and the result looks clean.

Still, you need to handle some more issues.

src/styles/stars.scss
edgergard: I guess so, if i set it only in stars__star block, my active star images would repeat 

image

src/index.html
```diff
@@ -4,9 +4,464 @@
...
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__logo">
+        <a href="#home">
+          <img src="./images/logo.png" alt="MOYO_logo">
+        </a>
+      </div>
```
Anna-Tkachenko, line 18: Do you need 3 wrappers for the logo? Check if you really need the first `div` one

src/index.html
```diff
@@ -4,9 +4,464 @@
...
+
+    <main class="main">
+      <section class="main__section catalog">
+        <article class="catalog__card card" data-qa="card">
+          <div class="card__top">
+            <img
+                src="images/imac.jpeg"
+                alt="APPLE A1419 iMac 27 Retina 5K Monoblock"
+                class="card__top-image"
+            >
```
Anna-Tkachenko, line 71: check correct indents for attributes

src/index.html
```diff
@@ -4,9 +4,464 @@
...
+          </div>
+
+          <div class="card__review">
+            <div class="card__stars stars stars--4">
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+              <div class="stars__star"></div>
+            </div>
```
Anna-Tkachenko, line 89: will be better to use semantic tags here

src/index.html
```diff
@@ -4,9 +4,464 @@
...
+              $2,199
+            </p>
+          </div>
+
+          <a
+            href="#buy"
+            data-qa="card-hover"
+            class="card__button-buy"
+          >
+            BUY
```
Anna-Tkachenko, line 111: Let's transform this text using css property

src/styles/card.scss
```diff
@@ -0,0 +1,110 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  box-sizing: border-box;
```
Anna-Tkachenko, line 5: will be better to add this property for all elements using `*` selector

src/styles/card.scss
```diff
@@ -0,0 +1,110 @@
+.card {
+  display: flex;
+  flex-direction: column;
```
Anna-Tkachenko, line 3: do you need these properties for the card? block elements are aligned vertically by default

Anna-Tkachenko (changes requested):

src/styles/_reset.scss
```diff
@@ -0,0 +1,16 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+h3 {
+  margin: 0;
+}
+
+p {
+  margin: 0;
```
Anna-Tkachenko, line 15: You can combine these styles using `,` for selectors

PR #3485, opened 2023-07-25
add task solution layout_catalog

thaiduk:


src/index.html
```diff
@@ -4,9 +4,461 @@
...
+  <body class="catalog">
+    <header class="header">
+      <a
+        href="#home"
+        class="header__logo"
+      >
+        <img
+          src="images/logo.png"
+          alt="logo"
+          >
```
BudnikOleksii, line 21:
```suggestion
        <img
          src="images/logo.png"
          alt="logo"
        >
```

src/index.html
```diff
@@ -4,9 +4,461 @@
...
+              href="#video"
+            >
+            video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="catalog__section">
```
BudnikOleksii, line 102: In my opinion, in this case, it should be "catalog" instead of the body element, as in the body we also have navigation and other staff.
And inside of main, you can create a section - "catalog__section"

src/styles/catalog_card.scss
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  &:hover {
+    transform: scale(1.2);
+
+    .card__title {
+      color: #34568b;
+    }
+  }
```
BudnikOleksii, line 101: Let's add a transition for this hover effect too.
You can use sass shorthand nesting here as well for .card__title.

src/styles/header.scss
```diff
@@ -0,0 +1,58 @@
...
+  align-items: center;
+
+  &__logo {
+    display: flex;
+  }
+}
+
+.nav {
+  &__list {
+    list-style: none;
```
BudnikOleksii, line 15: Let's create separate files for each BEM block.

src/styles/header.scss
```diff
@@ -0,0 +1,58 @@
...
+    font-size: 12px;
+    text-transform: uppercase;
+    cursor: pointer;
+    display: flex;
+    margin-left: 20px;
+    transition-duration: $animationTransitionDuration;
+
+    &--noMargin {
+      margin-left: 0;
+    }
```
BudnikOleksii, line 36: In this case, you have to specify in the html this class. You can set margins to "nav__list-item", and use the first or last child instead to get rid of the margin for the first element.

src/styles/normalizes.scss
```diff
@@ -0,0 +1,15 @@
+.catalog {
+  margin: 0;
+  font-family: "Roboto", sans-serif;
+  font-weight: 500;
```
BudnikOleksii, line 4: BEM block shouldn't have any margins. And as I mentioned before the body isn't a catalog in our case.

[STYLES] - Remember to use fallback fonts - alternative font-family in case the main one doesn't work [like this](https://www.w3schools.com/cssref/pr_font_font-family.asp)
image

src/styles/stars.scss
```diff
@@ -0,0 +1,19 @@
...
+  &__star {
+    width: 16px;
+    height: 16px;
+    margin-right: 4px;
+    background-position: center;
+    background-repeat: no-repeat;
+    background-image: url(../images/star.svg);
+    &:nth-child(-n+4) {
+      background-image: url(../images/star-active.svg);
+    }
```
BudnikOleksii, line 13:
```suggestion
    background-image: url(../images/star.svg);
    
    &:nth-child(-n+4) {
      background-image: url(../images/star-active.svg);
    }
```

BudnikOleksii (changes requested):
Looks great🔥
But could be even better, added a few comments, check them.
Also, the links for the demo and test report are broken, please, fix them so the interviewer doesn't need to change them manually to check your work.

src/styles/catalog_section.scss
```diff
@@ -0,0 +1,9 @@
+.catalog__section {
+  padding-top: 50px;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: 1fr 1fr;
```
BudnikOleksii, line 6: You don't need "grid-template-rows: 1fr 1fr;" as it could be more cards and rows

src/styles/normalizes.scss
```diff
@@ -0,0 +1,16 @@
+:root {
+  font-family: Roboto, Helvetica, sans-serif;
+  font-weight: 500;
+  margin: 0;
+}
```
BudnikOleksii, line 5: It's better apply to the body or html element, :root in most cases is used for css variables. Margin here is make no sense

BudnikOleksii (approved):
Good job👍
Left a few suggestions.

PR #3501, opened 2023-07-31
add catalog task solution

PodvAx:
- [DEMO LINK](https://PodvAx.github.io/layout_catalog/)
- [TEST REPORT LINK](https://PodvAx.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,12 +1,633 @@
+<!-- #region head -->
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    >
   </head>
-  <body>
-    <h1>Catalog</h1>
+  <!-- #endregion -->
+  <body class="page__body">
+    <!-- #region header -->
+    <header class="header page__header">
+      <a href="#home" class="header__link">
+        <img
+          src="./images/logo.png"
+          alt="Moyo logo"
+          class="logo"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="is-active nav__link">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#laptop_computers"
+              class="nav__link"
+              data-qa="nav-hover"
+            >
+              Laptops & Computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">Gadgets</a>
+          </li>
+
+          <li class="nav__item">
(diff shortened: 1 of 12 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  &__title {
+    color: $mainAccent;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+    margin-top: 40px;
+    margin-bottom: 4px;
```
denys-danyliuk, line 30: 5'th point in checklist

src/styles/blocks/main-content.scss
```diff
@@ -0,0 +1,35 @@
+@function columnsInLayout($numberOfColumns) {
+  @return repeat(#{$numberOfColumns}, #{$card-width});
+}
+
+.main-content {
+  display: grid;
+  place-items: center;
+  place-content: center;
+  grid-template-columns: columnsInLayout(1);
+  // grid-template-columns: repeat(4, #{$card-width});
```
denys-danyliuk, line 10: delete comments

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,117 @@
...
+    margin: 0 auto;
+    width: 160px;
+    height: 134px;
+
+    //scale image and add border to it on hover
+    transition: transform 0.4s, width 0.4s;
+
+    &:hover {
+      transform: scale(1.2);
+      border: 3px solid $grayElements;
```
denys-danyliuk, line 19: don't add new border on hover, instead add one with transparent border, and it's color on hover, also mention it in transition

denys-danyliuk (changes requested):

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,120 @@
+.card {
+  box-sizing: border-box;
+  width: $card-width;
+  border: 1px solid $grayElements;
+  border-radius: 5px;
+
+  &__image {
+    // display: block;
```
mykhalenych, line 8: delete all comments

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,120 @@
+.card {
+  box-sizing: border-box;
```
mykhalenych, line 2: move to `*` selector and delete in all other places

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,120 @@
...
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+    margin-top: 33px;
+
+    transition: color 0.3s;
+  }
+
+  &__info-text {
+    font-weight: 400;
```
mykhalenych, line 37: it is by default

src/index.html
```diff
@@ -1,12 +1,442 @@
...
+          src="./images/imac.jpeg"
+          alt="APPLE A1419 iMac 27&quot; Retina 5K Monoblock (MNED2UA/A)"
+          class="card__image"
+        >
+
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="card__info-text card__info-text--product-code">
```
mykhalenych, line 136: why do you use 3 classes? I think you can you only one

mykhalenych (changes requested):

PodvAx:
Where can be a problem? When I add a border of 3px to the picture, I understand that I should change width, height, paddings and margins, so the indents between elements can be perfect. But when I removed 3px from margin-top of card__title, tests did not passed, when i removed 6px it didn't passed, too. But when i removed 7px it passed. So what i do wrong?

src/index.html
PodvAx: Of course i can. But can i use element with modifier without element itself according to the BEM methodology? And there are only 2 classes.

Esceype (approved):
GJ!!!
About your question, you can remove border, make margin-top for card__title like in mockup and provide for img inline-paddings

PodvAx:
> GJ!!! About your question, you can remove border, make margin-top for card__title like in mockup and provide for img inline-paddings

Thanks! But I already had the image without border and added it on hover. But @denys-danyliuk said I don't need this, and i should add transparent border. So that's why I had this question.

PR #3510, opened 2023-08-03
Solution

EvgenBilenkiy:


src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,42 @@
...
+    display: flex;
+    align-items: center;
+
+    &:hover {
+      color: $blue-accent;
+    }
+  }
+}
+
+.is-active {
```
lerastarynets, line 28: The '.is-active' class should be a modifier of a specific element rather than a standalone class. This way, it will be more consistent with the BEM methodology.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,8 @@
+$font-color: #060b35;
+$blue-accent: #00acdc;
+$secondColor: #616070;
```
lerastarynets, line 3: use either kebab-case or camelCase for your variables

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #3584, opened 2023-08-30
catalog

liiasav:
- [DEMO LINK](https://liiasav.github.io/layout_catalog/)
- [TEST REPORT LINK](https://liiasav.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,9 +4,567 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    >
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com" crossorigin="anonymous"
+    >
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap" rel="stylesheet"
+    >
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="header__container">
+        <a href="#home" class="logo">
+          <img src="/images/logo.png" alt="logo">
+        </a>
+        <nav class="nav">
+          <ul class="nav__list-items">
+            <li class="nav__items">
+              <a
+                href="#apple"
+                class="nav__link is-active"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -4,9 +4,567 @@
...
+
+          <div class="stars">
+            <div class="stars__icon">
+              <div
+                class="stars__review"
+              >
+              </div>
+              <div
+                class="stars__review"
+              >
```
volodymyr-soltys97, line 97: When there are only 1 or 2 attributes, then you do not need to move them to a new line, you need to do this only when there are 3 or more attributes, you need to fix it everywhere
```suggestion
              <div class="stars__review">
```

src/styles/main.scss
```diff
@@ -1,3 +1,218 @@
...
+      width: 16px;
+      height: 16px;
+    }
+
+    .stars__review:not(:last-child) {
+      margin-right: 4px;
+      background-image: url(../images/star-active.svg);
+    }
+  }
+
```
volodymyr-soltys97, line 209: You can create a separate file for these styles

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/index.html
```diff
@@ -4,9 +4,462 @@
...
+            <div class="stars__icon">
+              <div class="stars__review"></div>
+              <div class="stars__review"></div>
+              <div class="stars__review"></div>
+              <div class="stars__review"></div>
+              <div class="stars__review"></div>
+            </div>
+            <div class="stars__count">
+              Reviews: 5
+            </div>
```
volodymyr-soltys97, line 198: Use a semantic tag everywhere for text

src/index.html
```diff
@@ -4,9 +4,462 @@
...
+              $2,199
+            </p>
+          </div>
+
+          <a
+            class="card__buy"
+            href="#"
+          >
+            Buy
+          </a>
```
volodymyr-soltys97, line 166: If we have only 2 attributes, then we don't need to move them to a new line, we only need to do this when there are 3 or more attributes, we need to fix it everywhere
```suggestion
          <a class="card__buy" href="#">
            Buy
          </a>
```

volodymyr-soltys97 (changes requested):

nazarmatsevych (approved):
👍

PR #3619, opened 2023-09-11
implement catalog v1

Drach-Illya:
- [DEMO LINK](https://Drach-Illya.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Drach-Illya.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,10 +3,421 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
+      crossorigin="anonymous"
+    >
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    >
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a href="#home" class="header__icon">
+        <img src="./images/logo.png" alt="Company Logo">
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#laptops & Computers"
+              class="nav__link"
+              data-qa="nav-hover"
+            >
+                Laptops & Computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">Gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#tablets" class="nav__link">Tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#photo" class="nav__link">Photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="card" data-qa="card">
+        <img
+          src="images/imac.jpeg"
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -3,10 +3,421 @@
...
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="card" data-qa="card">
```
buffi4ka, line 69: in my opinion it's better to create ul-list instead div. Semantic

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  @media (min-width: 768px) {
+    --column-number: 3;
+  }
+
+  @media (min-width: 1024px) {
+    --column-number: 4;
+  }
+
```
buffi4ka, line 22: **flex layout**: there should always be 4 cards in a row (not 2, 3 or 5)

src/styles/utilities/variables.scss
```diff
@@ -0,0 +1,8 @@
+$card-width: 200px;
+$secondary-font-size: 10px;
+$secondary-line-height: 14px;
+$main-accent: #060b35;
+$secondary: #616070;
+$white: #fff;
+$Blue-accent: #00acdc;
+$transition-duration: 300ms;
```
buffi4ka, line 8: [LAYOUT] - Use sass functions for dynamic max-width calculation, your catalog container should work fine for any number of cards. I didn't see this function

buffi4ka (approved):
Good job! You can make a little fixing according my comments. Good luck!

src/styles/blocks/container.scss
Drach-Illya: - ❗️❗️❗️ **MANDATORY if you're familiar with the `grid`**
  - use `grid` instead of `flex`
  - implement the [Grid Layout instead](./grid.md)
  
  # Catalog using grid layout
Implement the layout using `grid`
- Gap between cards should be: horizontal - `48px`, vertical - 46px
- Blocks should start from the left
- Media queries:
  - There should be four columns starting at `1024px`
  - There should be three columns starting at `768px`
  - There should be two columns starting at `488px`
  - There should be one column below `488px`
  
  As we are familiar with the Grid, we have to work with the rules given for grid, as the rule you noted is given when you construct the catalog using flex instead of grid.

src/index.html
Drach-Illya: Mentor mentioned that such a change would not be necessary to implement.

polosanya: I think section/article would be better here

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+    text-decoration: none;
+    margin-top: 16px;
+
+    border-radius: 5px;
+    background: $Blue-accent;
+    border: 1px none $Blue-accent;
+
+    &:hover {
+      background: none;
+      border-style: solid;
```
polosanya, line 89:
```suggestion
    border: 1px solid $Blue-accent;

    &:hover {
```
Just make the border solid by default and do nothing on hover. In that case, it will not affect other elements
hover

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
+.card {
+  box-sizing: border-box;
+  width: $card-width;
+  padding: 16px;
+
+  transition-duration: $transition-duration;
```
polosanya, line 6: It would be great to add transition-properties too
https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,94 @@
...
+
+  border-radius: 5px;
+  border: 1px solid #f3f3f3;
```
polosanya, line 9: Let's create variables for all colors

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,49 @@
+.nav {
+
+  @media (max-width: 850px) {
```
polosanya, line 3:
```suggestion
.nav {
  @media (max-width: 850px) {
```

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/styles/blocks/card.scss
Drach-Illya: How exactly would you recommend doing that? If I try to write transition: [property-name] $transition-duration; for Each relevant property, I recieve a mistake

polosanya (approved):
👍

PR #3650, opened 2023-09-12
add Catalog task solution

o-drozzdyk:
- [DEMO LINK](https://o-drozzdyk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://o-drozzdyk.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/product-cards.scss
+++ b/src/styles/blocks/product-cards.scss
@@ -1,0 +1,124 @@
+.main {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
+  gap: 46px 48px;
+
+  padding: 50px 40px;
+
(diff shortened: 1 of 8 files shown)


src/styles/blocks/product-cards.scss
```diff
@@ -0,0 +1,124 @@
...
+
+  &__photo {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin: 0 auto 40px;
+  }
+
+  &__title {
+    margin: 0 0 4px;
```
vasylmiskiv, line 44: You alredy reset your default style so you don't need to write 0 margins to top left right directions. Would be better to write `margin bottom`

vasylmiskiv (approved):

PR #3662, opened 2023-09-18
add task solution

pzwolenik:
- [DEMO LINK](https://pzwolenik.github.io/layout_catalog/)
- [TEST REPORT LINK](https://pzwolenik.github.io/layout_catalog/report/html_report/)

Hi there! One thing - I can't find a reason why these two tests didn't pass. I would like to ask for a hint - the nav works fine in other tests :)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,352 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&amp;display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a href="#home" class="logo">
+        <img
+          src="./images/logo.png"
+          alt="Logo"
+          class="logo__image"
+        >
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#apple"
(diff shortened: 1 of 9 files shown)


plodnikate (changes requested):
Hi, your nav tag has not the same wide as on the test, feel free to write in the c
Screenshot 2023-09-18 at 22 20 42
hat the next time you have questions

pzwolenik:
> Hi, your nav tag has not the same wide as on the test, feel free to write in the chat the next time you have questions Screenshot 2023-09-18 at 22 20 42

Hi! Thank you, I decided to changes paddings to margins in my nav :) 
If possible, can you give me any hint, whether I can still use padding? It would be helpful for the future :)

src/index.html
```diff
@@ -5,8 +5,352 @@
...
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#apple"
+              class="nav__link nav__link--is-active is-active"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
```
plodnikate, line 31: Add empty lines between multiline sibling blocks of HTML. But don't add empty lines between parent and child elements, fix it in all places, please.

plodnikate (changes requested):
Good job, just check and follow this approach ([css-rule](https://9elements.com/css-rule-order/)) and pay attention to the tip.

plodnikate (commented):
Read this article https://blog.hubspot.com/website/css-margin-vs-padding

src/index.html
pzwolenik: Hi! I was using this order: https://github.com/necolas/idiomatic-css#declaration-order (it was mentioned few tasks ago) :) But I changed it, thanks for the tip :)

Viktor-Kost (approved):

PR #3663, opened 2023-09-18
catalog solution task

skaletskyi:
[DEMO LINK](https://skaletskyi.github.io/layout_catalog/)
[TEST REPORT LINK](https://skaletskyi.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/header/header_style.scss
+++ b/src/styles/header/header_style.scss
@@ -1,0 +1,62 @@
+.page {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.page__body {
+  margin: 0;
+}
+
+.header {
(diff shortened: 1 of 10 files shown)


src/styles/header/header_style.scss
```diff
@@ -0,0 +1,62 @@
+.page {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", sans-serif;
+}
+
+.page__body {
+  margin: 0;
+}
```
maxim2310, line 9: margin would be better to reset in reset file, and font you can set as default in main.scss or create normalize file

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

PR #3732, opened 2023-10-17
task done

jacekkosiba:
My layout does not pass the tests, but to me it looks pretty identical to the reference on all the major web browsers :)

- [DEMO LINK](https://jacekkosiba.github.io/layout_catalog/)
- [TEST REPORT LINK](https://jacekkosiba.github.io/layout_catalog/report/html_report/)


choeqq (changes requested):
Your button is little big, this causes the whole layout to shift, which causes all of the tests fail
image
Also the menu items seem to be little stacked on each other 
image

I think it is just that, and all of the tests will pass 😄

jacekkosiba:
Thanks for the feedback! It actually helped me pass the test! Two of them still does not pass, but I assume this is a bug in the tests - the data-qa="nav-hover" element should in fact have underline on hover.

src/styles/blocks/_header.scss
```diff
@@ -0,0 +1,143 @@
...
+    &:hover::after,
+    &.is-active::after {
+      opacity: 1;
+    }
+  }
+
+  @media screen and (min-width: 768px) {
+
+    // placeholders for media
+    %flex-center1 {
```
JoyCoffeeAddict, line 103: I'm not a fun of using names like `%flex-center1`, it will become difficult to read and maintain in a bigger project.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,39 @@
...
+}
+
+%flex-align-stretch {
+  display: flex;
+  align-items: stretch;
+}
+
+%pointer {
+  cursor: pointer;
+  user-select: none;
```
JoyCoffeeAddict, line 38: You don't want to disable `user-select` on clickable elements. Some users may need that feature, always think about accessibility.

JoyCoffeeAddict (changes requested):
Links should not have a line below them on `:hover` 
image

Very good job overall

mbruhler (approved):
Approved.

PR #3755, opened 2023-10-20
add task solution

ruslanagnatenko:
- [DEMO LINK](https://ruslanagnatenko.github.io/layout_catalog/)
- [TEST REPORT LINK](https://ruslanagnatenko.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,11 +2,414 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Catalog</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap"
+      rel="stylesheet"
+    >
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="logo">
+        <a href="#home" class="logo__link">
+          <img
+            src="images/logo.png"
+            alt="logo"
+            class="logo__image"
+          >
+        </a>
+      </div>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a href="#apple" class="nav__link is-active">Apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#samsung" class="nav__link">Samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#smartphones" class="nav__link">Smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#laptops_computers"
+              class="nav__link"
+              data-qa="nav-hover"
+            >
+              Laptops & Computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#gadgets" class="nav__link">Gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#tablets" class="nav__link">Tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#photo" class="nav__link">Photo</a>
+          </li>
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main>
+      <div class="cards">
+        <div
+          class="card cards__card"
+          data-qa="card"
+        >
+          <img
+            src="images/imac.jpeg"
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -2,11 +2,414 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
BudnikOleksii, line 6: Keep your attributes correctly formatted
image

Fix in all places

src/index.html
```diff
@@ -2,11 +2,414 @@
...
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <div class="logo">
+        <a href="#home" class="logo__link">
+          <img
+            src="images/logo.png"
+            alt="logo"
```
BudnikOleksii, line 22: alt attribute should describe the image content
image

src/index.html
```diff
@@ -2,11 +2,414 @@
...
+
+          <li class="nav__item">
+            <a href="#video" class="nav__link">Video</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main>
+      <div class="cards">
```
BudnikOleksii, line 72: You can name it as a catalog to avoid misunderstanding

src/styles/card.scss
```diff
@@ -0,0 +1,101 @@
...
+  padding: 32px 16px 16px;
+
+  border-radius: 5px;
+  border: 1px solid $light-grey-color;
+  background: $white;
+
+  transition: transform $effect-duration;
+
+  &:hover {
+    @include largeAndChangeColorOnHover;
```
BudnikOleksii, line 15: Looks like on hover card should change only `transform`, you don't need to change `color` for card on hover

src/styles/card.scss
```diff
@@ -0,0 +1,101 @@
...
+    box-sizing: border-box;
+    display: flex;
+    align-items: center;
+    width: 100%;
+    height: 40px;
+
+    border: 1px solid $c-accent;
+    border-radius: 5px;
+    background: $c-accent;
+    @include hover (background-color, $white);
```
BudnikOleksii, line 84: 👍

src/styles/header.scss
```diff
@@ -0,0 +1,65 @@
...
+  &__link {
+    display: block;
+  }
+
+  &__image {
+    display: block;
+  }
+}
+
+.nav {
```
BudnikOleksii, line 23: Create a separate file for each BEM block

src/styles/mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+    #{$property}: #{$toValue};
+  }
+}
+
+@mixin colorHover ($property, $toValue) {
+  transition: #{$property} $effect-duration;
+  &:hover {
+    #{$property}: #{$toValue};
+  }
+}
```
BudnikOleksii, line 13: It's the same `hover` mixin, get rid of duplicates

src/styles/mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+
+@mixin largeAndChangeColorOnHover {
+  &:hover {
+    transform: scale(1.2);
+    z-index: 1;
+    transition: transform $effect-duration,
+      color $effect-duration;
```
BudnikOleksii, line 20: `transition` should be applied to the element, not just `:hover` selector. Also, in this mixin color change is missing.

Actually, I think this mixin is redundant and easier just manually set when you need to specify for a few properties as in most cases they are different

src/styles/mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+    --columns: 3;
+    column-gap: $column-gap;
+    row-gap: $row-gap;
+  }
+
+  @include onDesctop {
+    --columns: 4;
+    column-gap: $column-gap;
+    row-gap: $row-gap;
+  }
```
BudnikOleksii, line 63:
```suggestion
  column-gap: $column-gap;
  row-gap: $row-gap;
    
  @include onPhone {
    --columns: 2;
  }

  @include onTablet {
    --columns: 3;
  }

  @include onDesctop {
    --columns: 4;
  }
```
- Gaps are the same in all cases, so you don't need to specify for all media.
- You can use the shorthand property `gap` as well. https://developer.mozilla.org/ru/docs/Web/CSS/gap

src/styles/mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+@mixin secondary {
```
BudnikOleksii, line 66: secondary what?

src/styles/mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+@mixin card-title {
+  font-family: inherit;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 18px;
+  color: $main-text-color;
+}
```
BudnikOleksii, line 78: This mixin is redundant as you can use it only for the `card-title`

BudnikOleksii (changes requested):
Good job 👍 
Left some comments and suggestions. Check them

src/styles/header.scss
polosanya: Would be great to create a separate file for `nav` block too

polosanya (approved):
Well done! 🚀

PR #3758, opened 2023-10-20
add task solution

errisi:
- [DEMO LINK](https://errisi.github.io/layout_catalog/)
- [TEST REPORT LINK](https://errisi.github.io/layout_catalog/report/html_report/)

- [X] All component follow BEM and use SCSS
- [X] repaeted sizes and special colors are put to variables
- [X] cards are shown in 1, 2, 3 or 4 columns based on screen resolution

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,9 +4,280 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Catalog</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link href="https://fonts.googleapis.com/css2?family=Handjet:wght@200;400;700&family=Roboto:wght@500;700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a href="index.html" class="logo">
+        <img
+          src="images/logo.png"
+          alt="logo"
+          class="logo__img"
+        >
+      </a>
+      <nav class="nav">
+        <ul class="nav__links">
+          <li class="nav__item">
+            <a class="nav__link is-active" href="#">apple</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#">samsung</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#">smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              href="#"
+              data-qa="nav-hover">
+              Laptops & computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#">gadgets</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#">tablets</a>
+          </li>
+
+          <li class="nav__item">
+            <a class="nav__link" href="#">photo</a>
+          </li>
+
+          <li class="nav__item">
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -4,9 +4,280 @@
...
+            <a class="nav__link" href="#">smartphones</a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              class="nav__link"
+              href="#"
+              data-qa="nav-hover">
+              Laptops & computers
+            </a>
```
BudnikOleksii, line 40:
```suggestion
            <a
              class="nav__link"
              href="#"
              data-qa="nav-hover"
            >
              Laptops & computers
            </a>
```
Keep your attributes correctly formatted

src/index.html
```diff
@@ -4,9 +4,280 @@
...
+      <div class="catalog">
+        <div class="card" data-qa="card">
+          <img
+            class="card__img"
+            src="images/imac.jpeg"
+            alt="APPLE A1419 iMac 27 Retina"
+          >
+          <h1 class="card__title">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h1>
```
BudnikOleksii, line 72: It could be only one `h1` for all page, so in your case, it has to be `h2` as you have a lot of cards

src/index.html
```diff
@@ -4,9 +4,280 @@
...
+            <p class="card__rate-reviews">Reviews: 5</p>
+          </div>
+          <div class="card__price">
+            <p class="card__price-text">Price:</p>
+            <p class="card__price-amount">$2,199</p>
+          </div>
+          <button class="card__buy">Buy</button>
+        </div>
+
+      </div>
```
BudnikOleksii, line 280: Don't add empty lines between parent and child elements.

src/styles/card.scss
```diff
@@ -0,0 +1,103 @@
...
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 18px;
+      color: $main-color;
+    }
+  }
+
+  &__buy {
+    display: inline;
+    // display: block;
```
BudnikOleksii, line 83: It's better to do this button an `a`(link) element. And you don't need to change the display property to inline. Also, if it's a link you can open in a new tab the product that you want to check

src/styles/card.scss
```diff
@@ -0,0 +1,103 @@
...
+    padding: 12px 70px;
+    border-radius: 5px;
+    border: 0;
+    transition-duration: $transition-duration;
+  }
+
+  &__buy:hover {
+    background: none;
+    color: $blue-color;
+    border: $card-button-border;
```
BudnikOleksii, line 101: `border: 0;` it doesn't make sense. The border should be 1px transparent and then on hover changes the border color only

src/styles/catalog.scss
```diff
@@ -0,0 +1,17 @@
...
+  display: flex;
+  width: auto;
+  align-items: center;
+  justify-content: center;
+  padding: 50px 40px;
+}
+
+.catalog {
+  display: grid;
+  width: 944px;
```
BudnikOleksii, line 11: Why `width` hardcoded like this?

src/styles/header.scss
```diff
@@ -0,0 +1,73 @@
...
+    justify-content: center;
+  }
+}
+
+.logo__img {
+  display: block;
+  height: 40px;
+}
+
+.nav {
```
BudnikOleksii, line 22: Create separate file for each BEM block

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions to improve your solution

PR #3792, opened 2023-10-30
Develop

paviliong561:
[DEMO LINK](https://paviliong561.github.io/layout_catalog/)
[TEST REPORT LINK](https://paviliong561.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -1,0 +1,80 @@
+body {
+  padding: 0;
+  margin: 0;
+  border: 0;
+  --header-color: #00acdc;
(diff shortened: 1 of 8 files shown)


src/styles/header.scss
lerastarynets, line 1: why don't u use nesting in `header` block? pls nest all elements like in `cards` file

lerastarynets (changes requested):
Almost there, just one problem left)

lerastarynets (approved):
You're making excellent progress!

PR #3871, opened 2023-12-01
Develop catalog MOYO

VerJane23:
Considering that NPM and tests do not work correctly, I am sending you with the following comment:
Parcel did not allow (???) to register mixins for tablets and desktop in the catalog file, that’s why you see “media” in the catalog file. In addition, the names of mixins have been corrected at the request of the linter in the kebab case, and spaces between groups of variables have been removed. The horizontal (46) gaps between cards were reduced to the same value - nothing but a simple gap was perceived by the linter. In general, I crammed it in as best I could. 5:3 ✖ Expected shorthand property "gap" declaration-block-no-redundant-longhand-properties

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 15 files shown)


VerJane23:
Please consider it as it is, indicate the reasons and remedy. Thank you.

SanyaBratashchuk (approved):
LGTM!

PR #3904, opened 2023-12-06
add task solution

Ar4ik12:
- [DEMO LINK](https://Ar4ik12.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Ar4ik12.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/cards-block/card.scss
+++ b/src/styles/cards-block/card.scss
@@ -1,0 +1,137 @@
+.conatiner {
+  width: 1024px;
+  margin: 0 auto;
+  padding: 50px 40px;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(var(--column-count), 248px);
+}
+
+@media (min-width: 488px) {
+  .conatiner{
+    --column-count: 2;
+  }
+}
+
+@media (min-width: 768px) {
+  .conatiner{
+    --column-count: 3;
+  }
+}
+
+@media (min-width: 1024px) {
+  .conatiner{
+    --column-count: 4;
+  }
+}
+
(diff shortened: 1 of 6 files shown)


src/styles/cards-block/card.scss
```diff
@@ -0,0 +1,137 @@
+.conatiner {
```
l4st1m0za, line 1: There's a typo in the class name. It seems like it should be '.container' instead of '.conatiner'.

src/styles/cards-block/card.scss
```diff
@@ -0,0 +1,137 @@
...
+
+@media (min-width: 1200px) {
+  .conatiner{
+    width: 1200px;
+  }
+}
+
+.card  {
+  margin: 0 auto;
+  padding: 32px 16px 16px;
```
l4st1m0za, line 36: Margin is defined twice for the '.card' class. Consider removing one of them to avoid confusion.

src/styles/main.scss
```diff
@@ -1,3 +1,29 @@
...
+  --column-count: 1;
+}
+
+h2 {
+  margin: 0;
+}
+
+p {
+  margin: 0;
 }
```
l4st1m0za, line 25: You can make this thing smaller

l4st1m0za (changes requested):
GJ!

1) When user clikcks on the links in the header, the global link should have the same field ending on the **#apple** for example

2) Your "Buy" button animation should be smooth

3) When I hover over the "Buy" button elemens below it are pushed down, cahnge this

PR #3911, opened 2023-12-11
add task solution

sophiasph:
- [DEMO LINK](https://sophiasph.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sophiasph.github.io/layout_catalog/report/html_report/)


src/styles/cards.scss
```diff
@@ -0,0 +1,100 @@
+@import "./stars";
+
```
Esceype, line 2: Rename this file, if you card block inside so this file should call `card`.

src/styles/header.scss
```diff
@@ -0,0 +1,70 @@
...
+  padding: 0 50px;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+
+  &__logo {
+    width: 40px;
+    height: 40px;
+  }
+}
+
+.nav {
```
Esceype, line 14: This is separate block so move it to separate file.

src/styles/stars.scss
```diff
@@ -0,0 +1,21 @@
+.card__rating-stars {
```
Esceype, line 1: It looks like card element and if you create it to use BEM mix methodology then you need to keep this class in card file but make it separate from stars block

src/index.html
```diff
@@ -5,8 +5,407 @@
...
+        </ul>
+      </nav>
+    </header>
+
+    <main class="main">
+      <div class="card" data-qa="card">
+        <img
+          class="card__product-image"
+          src="./images/imac.jpeg"
+          alt="Imac"
```
Esceype, line 100: Provide better naming for alt attribute

Esceype (changes requested):
GJ!
Read the following comments

src/index.html
sophiasph: why is it alt attribute name bad?

maxim2310 (changes requested):
something went wrong with your demo link, plz fix it 
image

PR #3974, opened 2024-01-08
layout_catalog 'add solution'

Dandriichuk:
- [DEMO LINK](https://Dandriichuk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Dandriichuk.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/block-card/card.scss
+++ b/src/styles/block-card/card.scss
@@ -1,0 +1,111 @@
+.card {
+  max-width: 200px;
+  border-radius: 5px;
+  border: 1px solid $background-color;
+
+  &:hover {
+    transform: scale(1.2);
+    transition: ease 300ms;
+  }
+
+  &__img {
+    display: flex;
+    justify-content: center;
+    padding: 32px 19px 0;
+    margin-bottom: 40px;
+
+    &-link {
+      height: 134px;
+    }
+  }
+
+  &__tittle {
+    margin-bottom: 4px;
+    padding-inline: 16px;
+    color: $main-accent;
+    font-size: 12px;
+    font-weight: 500;
+    font-style: normal;
+    line-height: 18px;
+
+    &:hover {
+      color: #34568b;
+      transition: ease 300ms;
+    }
+  }
+
+  &__code {
+    margin-bottom: 16px;
+    padding-inline: 16px;
+    color: $secondary-accent;
+    font-size: 10px;
+    font-weight: 400;
+    line-height: 14px;
+  }
+
+  &__reviews {
+    display: flex;
+    align-items: flex-end;
+    justify-content: space-between;
+    margin-bottom: 24px;
+    padding-inline: 16px;
+
+    &-text {
+      color: $main-accent;
+      font-size: 10px;
+      font-weight: 400;
+      font-style: normal;
+      line-height: 14px;
+      text-align: right;
+    }
+  }
+
+  &__price {
+    display: flex;
+    text-align: center;
+    justify-content: space-between;
+    margin-bottom: 16px;
+    padding-inline: 16px;
+
+    &-text {
+      color: $secondary-accent;
+      font-size: 12px;
+      font-weight: 400;
+      line-height: 18px;
+    }
+
+    &-value {
+      color: $main-accent;
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 18px;
+    }
+  }
+
+  &__button {
+    display: flex;
+    margin: 16px;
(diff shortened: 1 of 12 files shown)


src/styles/block-card/card.scss
```diff
@@ -0,0 +1,111 @@
...
+    padding-inline: 16px;
+    color: $main-accent;
+    font-size: 12px;
+    font-weight: 500;
+    font-style: normal;
+    line-height: 18px;
+
+    &:hover {
+      color: #34568b;
+      transition: ease 300ms;
```
anastasiia-tilikina, line 33: transition should be set on default class not on :hover pseudo-class, otherwise it'll work wrong.

src/styles/block-card/card.scss
```diff
@@ -0,0 +1,111 @@
...
+    margin-bottom: 24px;
+    padding-inline: 16px;
+
+    &-text {
+      color: $main-accent;
+      font-size: 10px;
+      font-weight: 400;
+      font-style: normal;
+      line-height: 14px;
+      text-align: right;
```
anastasiia-tilikina, line 59: no need to set it here. You've already set justify-content: space-between.  that's enough here

src/styles/block-card/card.scss
```diff
@@ -0,0 +1,111 @@
...
+      text-decoration: none;
+      color: $background-color;
+      text-align: center;
+
+      &:hover {
+        border: 1px solid $button-accent;
+        background-color: $background-color;
+        color: $button-accent;
+        border-radius: 5px;
+        transition: ease 300ms;
```
anastasiia-tilikina, line 107: same here with transition.

src/styles/block-card/stars.scss
```diff
@@ -0,0 +1,20 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    height: 16px;
+    width: 16px;
+
+    &:not(:last-child) {
+      margin-right: 4px;
+    }
+
+    &:nth-child(-n + 4) {
```
anastasiia-tilikina, line 16: You should add here all modifiers for stars from there previous task

src/styles/block-card/card.scss
```diff
@@ -0,0 +1,111 @@
...
+      line-height: 16px;
+      font-size: 14px;
+      font-weight: 700;
+      text-transform: uppercase;
+      text-decoration: none;
+      color: $background-color;
+      text-align: center;
+
+      &:hover {
+        border: 1px solid $button-accent;
```
anastasiia-tilikina, line 103: It is always good to set border by default and change only color on hover

src/styles/block-card/card.scss
anastasiia-tilikina: Also set properties for transition explictly or use all key word and check [transition](https://developer.mozilla.org/en-US/docs/Web/CSS/transition) shorthand order

src/styles/block-header/nav.scss
```diff
@@ -0,0 +1,58 @@
+.nav {
+  display: flex;
+  margin: 0 50px 0 0;
```
anastasiia-tilikina, line 3: it is better to set padding to header instead of setting margin here. especially considering the fact that we shouldn't set outer positioning for BEM block.

src/styles/block-header/header.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  height: 60px;
+  margin: 0;
+  width: auto;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+
+  &__logo {
+    margin-left: 50px;
```
anastasiia-tilikina, line 14: same here.

src/styles/block-header/nav.scss
```diff
@@ -0,0 +1,58 @@
...
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  &__item {
+    margin: 0 0 0 20px;
```
anastasiia-tilikina, line 15:
```suggestion
    margin-left: 20px;
```

anastasiia-tilikina (changes requested):
Good job!
check my commetns.

Pay attention to deleting font-style and proper setting transition everywhere in your project

maxim2310 (changes requested):
Plz fix all previous comments

volodymyr-soltys97 (approved):
Great work 🔥

PR #4195, opened 2024-03-10
Develop

vinasxz:
- [DEMO LINK](https://vinasxz.github.io/layout_catalog/)
- [TEST REPORT LINK](https://vinasxz.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,7 +9,7 @@ Create HTML page with catalog. Develop semantic page structure as shown on [the 
 - add `data-qa="card-hover"` (not just `hover`) to the link `Buy` inside the first card
 - nav links color is not `black` any more (nav links should have `#060b35` color)
 - add class `is-active` to the first link (`Apple`) in navigation
-- use `<main>` tag for cards container 
+- use `<main>` tag for cards container
 - use grid for cards with different number of columns:
   - 1 for the smaller screens
   - 2 starting at `488px`
@@ -30,14 +30,14 @@ Make all the changes smooth on hover (during 300ms):
 ## Checklist
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_catalog/report/html_report/)
+- [DEMO LINK](https://vinasxz.github.io/layout_catalog/)
+- [TEST REPORT LINK](https://vinasxz.github.io/layout_catalog/report/html_report/)
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
-- [ ] All component follow BEM and use SCSS
-- [ ] repaeted sizes and special colors are put to variables
-- [ ] Grid is used for the columns
-- [ ] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
-- [ ] All changes on `:hover` are smooth
-- [ ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)
+- [x] All component follow BEM and use SCSS
+- [x] repaeted sizes and special colors are put to variables
+- [x] Grid is used for the columns
+- [x] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
+- [x] All changes on `:hover` are smooth
+- [x] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -1,0 +1,125 @@
+.catalog {
+  box-sizing: border-box;
+  display: grid;
+  grid-gap: 46px 48px;
+  justify-content: center;
+  grid-template-columns: repeat($column-count, 200px);
+  margin: 0 auto;
+  padding: 50px 0;
+
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, 200px);
+  }
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
+
+  &__card {
+    box-sizing: border-box;
+    display: flex;
+    flex-direction: column;
+    width: 200px;
+    border-radius: 5px;
+    border: 1px solid $primary-bg-color;
+    background-color: $background-card-color;
+
+    &:hover {
+      transform: scale(1.2);
+      transition: all 300ms ease;
+    }
+
+    &-image {
+      width: 160px;
+      height: 134px;
+      margin-top: 32px;
+      margin-right: 19px;
+      margin-left: 19px;
+    }
+
+    &-info {
+      margin: 40px 16px 0;
+
+      &-text {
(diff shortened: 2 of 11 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,126 @@
+.catalog {
+  box-sizing: border-box;
+  display: grid;
+  grid-gap: 46px 48px;
+  justify-content: center;
+  grid-template-columns: repeat($column-count, 200px);
```
pedro-ruas, line 6: This line isn't being used.
Just below it, you're declaring your media queries, and there are declarations for any width, thus those declarations are overwritting this one here.
The best approach you can do here is to remove the variable $column-count and replace it with `1`, then you can also remove the declaration for `@media (max-width: 487px)`.
The rest can remain the same

src/styles/card.scss
```diff
@@ -0,0 +1,126 @@
...
+    display: flex;
+    flex-direction: column;
+    width: 200px;
+    border-radius: 5px;
+    border: 1px solid $primary-bg-color;
+    background-color: $background-card-color;
+
+    &:hover {
+      transform: scale(1.2);
+      transition: all 300ms ease;
```
pedro-ruas, line 37: This transition should be on the element without `:hover`.
By adding the transition only to the :hover state, what happens is that the transition applies when you start hovering, but not when you stop hovering, because at that moment the element is not in the :hover state anymore.
This creates a behavior where it is smooth to  check the item, but abrupt the moment you stop checking it.
To fix, just move `transition: all 300ms ease` inside `&__card` and outside `:hover`

src/styles/card.scss
```diff
@@ -0,0 +1,126 @@
...
+        line-height: 16px;
+        font-size: 14px;
+      }
+
+      &:hover {
+        border-radius: 5px;
+        border: 1px solid $button-main-color;
+        background-color: $background-card-color;
+        color: $button-main-color;
+        transition: all 300ms ease;
```
pedro-ruas, line 122: This is the same issue I've pointed in the previous comment, you can fix it by moving `transition` property to `button` directly

src/styles/stars.scss
```diff
@@ -0,0 +1,38 @@
...
+        width: 16px;
+        height: 16px;
+        margin-right: 4px;
+        background: url(../images/star.svg) no-repeat center;
+
+        &:last-child {
+          margin-right: 0;
+        }
+
+        &:nth-child(-n + 4) {
```
pedro-ruas, line 23: By using this styling, you are fixing the cards to always be 4 stars active and 1 inactive.
A better approach would be to implement the Stars block the same way you did before on the Stars task.
With that implementation, the modifier for --0, --1... --5 is present at `.stars` class and this changes the actual stars count.
I believe it's a google practice to implement it here and you can do it quite easily by using your own code from Stars as an example ;)

hint: you probably need to check your html too

src/styles/vars.scss
```diff
@@ -0,0 +1,7 @@
+$primary-bg-color: #f3f3f3;
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+$button-main-color: #00acdc;
+$background-card-color: #fff;
+$hover-color: #34568b;
+$column-count: 1;
```
pedro-ruas, line 7: This variable will not be necessary anymore after the fixes from previous comments, you can delete it :)

src/styles/vars.scss
```diff
@@ -0,0 +1,7 @@
+$primary-bg-color: #f3f3f3;
+$main-text-color: #060b35;
+$secondary-text-color: #616070;
+$button-main-color: #00acdc;
+$background-card-color: #fff;
+$hover-color: #34568b;
```
pedro-ruas, line 6: I see you did not use this variable anywhere in the code.
However, you should use it accordingly to [this part of the Checklist](https://github.com/mate-academy/layout_catalog#:~:text=change%20the%20card%20title%20text%20color%20to%20%2334568b%20when%20the%20card%20is%20hovered%20(.card%3Ahover%20.card__title))

pedro-ruas (changes requested):
Vinicius, your code looks great for the most part.
There is 2 animation issues (actually, it's one issue that happens twice)
And you missed 1 item on  the checklist, but easy fix.

readme.md
```diff
@@ -9,7 +9,7 @@ Create HTML page with catalog. Develop semantic page structure as shown on [the
 - add `data-qa="card-hover"` (not just `hover`) to the link `Buy` inside the first card
 - nav links color is not `black` any more (nav links should have `#060b35` color)
 - add class `is-active` to the first link (`Apple`) in navigation
-- use `<main>` tag for cards container 
+- use `<main>` tag for cards container
```
mgrinko, line 12:
```suggestion
- use `<main>` tag for cards container 
```

src/styles/stars.scss
EdPirro, line 1: Your stars implementation should follow the previous exercise [Stars](https://github.com/mate-academy/layout_stars) but converted to scss [Referece](https://github.com/mate-academy/layout_catalog#:~:text=use%20Header%2C%20Stars%20and%20Card%20blocks%20from%20previous%20tasks%20but%20rewrite%20them%20using%20BEM%20and%20SCSS)

EdPirro (changes requested):
Good effort, @vinasxz! 👏 You're almost there, please thoroughly review the requested changes, if you have any questions please ask them in the qna channel!

src/styles/stars.scss
EdPirro: @vinasxz please have a look a this, it is still unresolved!

src/styles/card.scss
```diff
@@ -0,0 +1,126 @@
...
+      &-text {
+        font-size: 12px;
+        font-weight: 500;
+        line-height: 18px;
+        margin: 0;
+        color: $main-text-color;
+        font-style: normal;
+        &-text:hover {
+          color: $hover-color;
+        }
```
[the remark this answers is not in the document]
vinasxz: @EdPirro Yes, it should. My mistake was that when I was trying to fix the hover effect, I didn't place it in the correct paragraph. Thanks for that.

src/styles/stars.scss
```diff
@@ -0,0 +1,38 @@
...
+
+    &-list {
+      display: flex;
+
+      &-star {
+        width: 16px;
+        height: 16px;
+        margin-right: 4px;
+
+        @for $index from 1 through 4 {
```
pedro-ruas, line 18: Vinicius, I see this change was due to our last chat, and I may have failed in my explanation.
The idea of using a @for would be to  make possible to change the stars count by just using a modifier at `card-stars-list` because that is what packs all stars.
The way you implemented maintains the same results than the previous iteration, always keeping 4 stars active and one inactive.

Here is my suggestion to fix the problem:
 1 - In your HTML, when you call `catalog__card-stars-list` also add a modifier, default it at 4, like so: `class="catalog__card-stars-list catalog__card-stars-list--4`, and each `star` shouldn't have a modifier. Remove "-1, -2, -3, -4, -5"
 
2 - this code here should be moved to &-list, in this way:
```
.catalog {
  &__card-stars {
    ...
    &-list {
      ...
      @for $index from 1 through 5 {
        &--#{$index} {
          .catalog__card-stars-list-star:nth-child(-n+#{$index}) {
             background: url(../images/star-active.svg) no-repeat center;
          }
        }
      }

      &-star {
        ...
        background: url(../images/star.svg) no-repeat center;
      }
    }
  }
}
```

In this way, the modifier added to the list will alter all stars to match the correct styling

pedro-ruas (changes requested):

EdPirro (approved):
Good job, @vinasxz! Your solution looks great and fulfills all the requirements. Keep it up! 🚀

PR #4244, opened 2024-03-18
add task solution

SlawikH:
- [DEMO LINK](https://SlawikH.github.io/layout_catalog/)
- [TEST REPORT LINK](https://SlawikH.github.io/layout_catalog/report/html_report/)

Made some changes to my previous code from tasks and improved it
Looking forward for your comments and remarks
Please, let me know if something wrong :)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,6 +22,364 @@
   </head>
 
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a
+        href="#"
+        class="header__logo"
+      >
+        <img
+          src="/src/images/logo.png"
+          alt="MOYO logo"
+        />
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link is-active"
+            >
+              Apple
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+            >
+              Samsung
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+            >
+              Smartphones
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+              data-qa="nav-hover"
+            >
+              Laptops & Computers
+            </a>
+          </li>
+
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+            >
+              Gadgets
+            </a>
+          </li>
+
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -22,6 +22,364 @@
...
+        />
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="card__code">Product code: 195434</p>
+
+        <div class="card__rate">
+          <div class="stars stars--4">
+            <div class="stars__star"></div>
```
DanilWeda, line 132: Try to use ul and li in this case.
it's preferred because we have one repeated element inside div.
In this case best variant use list)

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,25 @@
...
+
+    margin-right: 4px;
+  }
+}
+
+.stars--5 .stars__star:nth-child(-n + 5),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--1 .stars__star:nth-child(-n + 1) {
```
DanilWeda, line 23: You can write more universal variant of this ( using SCSS capabilities)

DanilWeda (changes requested):
Great job! 
Check some small comments. 
Also, check a transition ( hover has a transition, unhover - doesn`t have a transition).
Move transition to element, not use inside `:hover`. 
Add cursor pointer to action elements  (buttons).
Many thanks)

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,63 @@
...
+  bottom: 0;
+  left: 0;
+
+  height: 4px;
+  width: 100%;
+
+  background-color: $color-blue-accent;
+
+  border-radius: 8px;
+}
```
alexander-ignatow, line 63: Are you sure this needs to be a separate block? Maybe it makes sense to make this modifier?

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  &__logo {
+    align-self: center;
+    justify-content: center;
+
+    height: 40px;
+  }
+}
+
+.nav {
```
alexander-ignatow, line 16: Separate BEM blocks should live in separate files

alexander-ignatow (approved):
Looks good - let's go with it 👍 

However, please mind my comments on BEM structure and file structure ☝️

alexander-ignatow (approved):

PR #4248, opened 2024-03-18
add solution

stepanyukvova200:
- [DEMO LINK](https://stepanyukvova200.github.io/layout_catalog/)
- [TEST REPORT LINK](https://stepanyukvova200.github.io/layout_catalog/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,121 @@
...
+    transition-duration: $hover-animation-duration;
+
+    &:hover {
+      color: $color__button--1;
+      background-color: $color__button--2;
+    }
+  }
+}
+
+.cards-container {
```
alexander-ignatow, line 100: This is different BEM block, so please move it to another file

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,66 @@
...
+  padding-right: 50px;
+}
+
+.logo {
+  display: inline-block;
+  align-self: center;
+  width: 40px;
+  height: 40px;
+  margin-left: 50px;
+}
```
alexander-ignatow, line 16: If you make it a separate block, please remove the margin and take it to the separate file
Or you can refactor it to the element (say `header__logo`)

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.nav {
```
alexander-ignatow, line 18: Separate block = separate file

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,66 @@
...
+.is-active::after {
+  content: '';
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 4px;
+  background-color: $color__active-link;
+  border-radius: 8px;
+}
```
alexander-ignatow, line 66: 1. Use nesting
2. Are you sure you need it as a block?

alexander-ignatow (changes requested):
It works, that is good

Please fix the comments below
Also, make sure to fix the tests

One more thing: make sure not to submit unnecessary files (in your case .css and .css.map). No need to fix now, but look for it in the future.

src/styles/blocks/header.scss
stepanyukvova200: yes, i need to have exectly class "is-active" for tests ( So i just moved it to separate file.

alexander-ignatow: > yes, i need to have exectly class "is-active" for tests ( So i just moved it to separate file.

Agree here - let's keep it

alexander-ignatow (approved):
Looks clean 👍 . Good job!

JFYI
On `is-active` class - probably the perfect solution would be following:
- Create modifier `blabla--is-active` and move all styling there. 
- Keep `is-active` class in HTML markup only for tests.

In such way the tests will work and the styling would be "more BEM"

PR #4258, opened 2024-03-21
add task solution

tarassamchenko:
- [DEMO LINK](https://tarassamchenko.github.io/layout_catalog/)
- [TEST REPORT LINK](https://tarassamchenko.github.io/layout_catalog/report/html_report/)
- [x ] All component follow BEM and use SCSS
- [x ] repaeted sizes and special colors are put to variables
- [x ] Grid is used for the columns
- [x ] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
- [x ] All changes on `:hover` are smooth
- [x ] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -1,0 +1,59 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  min-height: 60px;
+}
+
+.logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+  margin-left: 50px;
+}
+
+.nav {
+  display: flex;
+  margin-right: 50px;
+
+  &__list {
+    display: flex;
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    font-size: 12px;
+    text-decoration: none;
+  }
+
+  &__link {
+    display: inline-block;
+    line-height: 60px;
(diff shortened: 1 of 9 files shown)


src/styles/header.scss
```diff
@@ -0,0 +1,59 @@
...
+  justify-content: space-between;
+  align-items: center;
+  min-height: 60px;
+}
+
+.logo {
+  display: block;
+  width: 40px;
+  height: 40px;
+  margin-left: 50px;
```
alexander-ignatow, line 12: No margins for BEM blocks

DanilWeda (changes requested):
Good job. Check comments about margins.
Additionally, move the transition to the element. 
Now when you hover over an element I see a transition, but if I move the cursor outside an element, we do not have a transition. Make this step for better smoothing on the page. 
Many thanks)

alexander-ignatow (approved):
Good job - let's proceed with it 👍 

However - I still have few comments:
- File `header.scss` still can be improved - it contains multiple BEM blocks, while you should have 1 BEM block per one file
- Please mind your commit messages - `add better version` is not informative enough. Try describing the changes you made: `removed margins from BEM blocks`

PR #4271, opened 2024-03-27
add task solution

igorpolianskyi:
- [DEMO LINK](https://igorpolianskyi.github.io/layout_catalog/)
- [TEST REPORT LINK](https://igorpolianskyi.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,100 @@
+.card {
+  box-sizing: border-box;
+  width: $cardwidth;
+  height: 408px;
+  background-color: $base-color;
+  border: 1px solid $base-border-color;
+  border-radius: 5px;
+  display: flex;
+  flex-direction: column;
+  padding: 32px 16px 16px;
+  transition-duration: $duration;
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
+  &__image {
+    height: 134px;
+    width: 160px;
+    align-self: center;
+    margin-bottom: 40px;
+  }
+
+  &__code {
+    width: fit-content;
+    font-size: 10px;
+    line-height: 14px;
+    font-weight: 400;
+    color: $sub-text-color;
+    margin-bottom: 16px;
+  }
+
+  &__raiting {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    align-self: last baseline;
+    font-size: 10px;
+    line-height: 14px;
+    font-weight: 400;
+    color: $main-text-color;
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    line-height: 18px;
+    margin-bottom: 16px;
+  }
+
+  &__price-title {
+    font-size: 12px;
+    font-weight: 400;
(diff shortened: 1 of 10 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,100 @@
+.card {
+  box-sizing: border-box;
```
anastasiia-tilikina, line 2: it is better to set this property for all elements via * selector

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,100 @@
...
+  width: $cardwidth;
+  height: 408px;
```
anastasiia-tilikina, line 4: do not hardcode card height. It is not a good idea to hardcode elements with content inside

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,100 @@
...
+    justify-content: center;
+    box-sizing: border-box;
+    align-items: center;
+    color: $base-color;
+
+    font-size: 14px;
+    font-weight: 400;
+    line-height: 16px;
+
+    height: 40px;
```
anastasiia-tilikina, line 77: do not hardcode this either use proper padding instead

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
...
+@import 'blocks/logo';
+@import 'blocks/nav';
+
+* {
   margin: 0;
+  padding: 0;
+}
+
+html {
+  /* stylelint-disable-next-line font-family-name-quotes */
```
anastasiia-tilikina, line 15: why do you need disable linter here?

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
-body {
+@import 'blocks/var';
+@import 'blocks/catalog';
+@import 'blocks/header';
+@import 'blocks/stars';
+@import 'blocks/card';
+@import 'blocks/logo';
+@import 'blocks/nav';
+
+* {
```
anastasiia-tilikina, line 9: it is not a good practice to use * selector for any thing but border-box property. You can reset margins and padding for particular elements. for instance:

```scss
p,
h1,
ul {
  margin: 0;
  padding: 0;
}
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,100 @@
...
+    color: $sub-text-color;
+  }
+
+  &__price-value {
+    font-size: 16px;
+    font-weight: 700;
+    color: $main-text-color;
+  }
+
+  &__button {
```
anastasiia-tilikina, line 66: The text in the button should be upper case.
use text-transform property

anastasiia-tilikina (changes requested):
Very well done!
Very good structured code 👍🏼 

Let's just fix some issues here

src/styles/index.scss
igorpolianskyi: cause i set 'roboto' font-style here

vadiimvooo: You can just remove quotes for this font name.

vadiimvooo (approved):
LGTM

PR #4294, opened 2024-04-10
solutionV1

kgGugala:
- [DEMO LINK](https://kgGugala.github.io/layout_catalog/)
- [TEST REPORT LINK](https://kgGugala.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,214 @@
+$primary-color: #00acdc;
+$secondary-color: #616070;
+$tertiary-color: #060b35;
+$font-family: 'Roboto', sans-serif;
+
 body {
   margin: 0;
+  font-family: $font-family;
+}
+
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  background: #fff;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo {
+  display: block;
+  height: 40px;
+  width: 40px;
+
+  &__image {
+    height: 100%;
+    width: 100%;
+  }
+}
+
+.nav__list {
+  display: flex;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+}
+
+.nav__item {
+  margin-left: 20px;
+
+  &:first-child {
+    margin-left: 0;
+  }
+}
+
+.nav__link {
+  position: relative;
+  display: flex;
+  align-items: center;
+  height: 60px;
+  font-family: $font-family;
+  font-size: 12px;
+  font-weight: 500;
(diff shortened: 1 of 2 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,221 @@
...
+
+.nav__list {
+  display: flex;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+}
+
+.nav__item {
+  margin-left: 20px;
```
yevhenii-pyl, line 39: I'd prefer to use gap property whenever it's possible instead of margins.
As a general rule of thumb, you always want a parent component being responsible for child elements positioning.

src/styles/index.scss
```diff
@@ -1,3 +1,221 @@
...
+      border-radius: 2px;
+      background-color: $primary-color;
+    }
+  }
+}
+
+.main {
+  width: 100%;
+  display: flex;
+  justify-content: center;
```
yevhenii-pyl, line 83: `flex-wrap `may help here to make it a bit more responsive (it's not required here, but still)

src/styles/index.scss
```diff
@@ -1,3 +1,221 @@
...
+    line-height: 18px;
+
+    &:hover {
+      color: #34568b;
+    }
+  }
+
+  &__product-code {
+    color: $secondary-color;
+    margin: 4px 0 16px;
```
yevhenii-pyl, line 173: You tend to use margin-bottom here to position elements within the card. 
It's a valid approach, but let's keep it consistent by adding margin-bottom to the `card__name` element instead of setting top margin here.

src/styles/index.scss
```diff
@@ -1,3 +1,221 @@
...
+  }
+}
+
+.main {
+  width: 100%;
+  display: flex;
+  justify-content: center;
+}
+
+.main__catalog {
```
yevhenii-pyl, line 86: Let's use it as a BEM block:

&__catalog

yevhenii-pyl (changes requested):
Good job! 👍 
Let's just adjust a few tiny things and we are good to go.

And btw, could you double-check that the newest version of demo is deployed? Had to open your PR locally to see changes, demo looks a bit outdated.

src/styles/index.scss
```diff
@@ -1,3 +1,220 @@
...
+  display: flex;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  gap: 20px;
+}
+
+.nav__item {
+  &:first-child {
+    margin-left: 0;
```
yevhenii-pyl, line 41: Pretty much redundant here after removing margin-left from nav items :)

yevhenii-pyl (approved):
You got the idea 🚀

PR #4299, opened 2024-04-11
add task solution

szmyt421:
- [DEMO LINK](https://szmyt421.github.io/layout_catalog/)
- [TEST REPORT LINK](https://szmyt421.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -19,7 +19,7 @@
   &__image {
     width: 160px;
     height: 134px;
-    margin: 32px 19px 0;
+    padding: 20px;
   }
 
   &__content {
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -1,0 +1,99 @@
+@import './variables';
+
+.card {
+  width: 200px;
+  flex-shrink: 0;
+  background-color: $white;
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid $gray-elements;
+  display: flex;
+  flex-direction: column;
+
+  transition: all 300ms ease-in-out;
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    margin: 32px 19px 0;
+  }
+
+  &__content {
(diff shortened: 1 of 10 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,99 @@
...
+  flex-shrink: 0;
+  background-color: $white;
+  box-sizing: border-box;
+  border-radius: 5px;
+  border: 1px solid $gray-elements;
+  display: flex;
+  flex-direction: column;
+
+  transition: all 300ms ease-in-out;
+
```
yevhenii-pyl, line 14: Can't agree here: instead of having margins for image && content you should use padding for the card. 

Same here: parent is responsible for positioning its children.

src/styles/card.scss
```diff
@@ -0,0 +1,99 @@
...
+    color: $title-hover-color;
+  }
+
+  &__code {
+    color: $secondary;
+    font-size: 10px;
+    font-weight: 400;
+    line-height: 14px;
+    text-align: left;
+    margin-top: 4px;
```
yevhenii-pyl, line 46: 👍 

You started using margin-top to position elements and continued doing so. 
You can't use gap here since there is different spacing between elements, so using margin is a valid approach. 

Margin-top property is used consistently and never mixed with margin-bottom. Good job.

yevhenii-pyl (changes requested):
Solid work!
Just a few minor adjustments and we are good to go 👍

szmyt421:
- [DEMO LINK](https://szmyt421.github.io/layout_catalog/)
- [TEST REPORT LINK](https://szmyt421.github.io/layout_catalog/report/html_report/)

natalia-klonowska (changes requested):
you need to resolve all previous comments 
if you have problem with fixing them you can ask question in fx_chat or sing up for QnA (today is at 19:30)

natalia-klonowska (changes requested):
can you update your demo and test report links? Because on your screenshot I don't see any information to be able to help

PR #4325, opened 2024-04-22
Develop

jooaasss:
- [DEMO LINK](https://jooaasss.github.io/layout_catalog/)
- [TEST REPORT LINK](https://jooaasss.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/header/import_header.scss
+++ b/src/styles/blocks/header/import_header.scss
@@ -1,6 +1,6 @@
 body {
   $blue-color: #00acdc;
-  $black-color: #000;
+  $black-color: #060b35;
   $grey-color: #585858;
 
   font-size: 12px;
--- a/src/styles/blocks/header/import_header.scss
+++ b/src/styles/blocks/header/import_header.scss
@@ -1,0 +1,81 @@
+body {
+  $blue-color: #00acdc;
+  $black-color: #000;
+  $grey-color: #585858;
+
+  font-size: 12px;
+  font-weight: 500;
+
+  .header {
+    padding: 0 50px;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    color: #fff;
+    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+    &__logo {
+      display: block;
+      width: 40px;
+      height: 40px;
+
+      &-image {
+        width: 100%;
+        height: 100%;
+      }
+    }
+  }
+
+  .nav {
+    display: block;
+
+    &__list {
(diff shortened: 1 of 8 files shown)


src/styles/blocks/header/import_header.scss
```diff
@@ -0,0 +1,77 @@
+body {
```
TarasHoliuk, line 1: it's not part of `header` BEM block, so shouldn't be here
it should be defined in the `normalize.scss` / `reset.scss` or at least in the index.scss

src/styles/blocks/header/import_header.scss
```diff
@@ -0,0 +1,77 @@
...
+      height: 40px;
+
+      &-image {
+        width: 100%;
+        height: 100%;
+      }
+    }
+  }
+
+  .nav {
```
TarasHoliuk, line 26: same - it's a separate BEM block - should be in a separate file

src/styles/blocks/header/import_header.scss
```diff
@@ -0,0 +1,77 @@
...
+      color: #060b35;
+      height: 60px;
+      line-height: 60px;
+      text-decoration: none;
+      white-space: nowrap;
+
+      transition: color 0.3s;
+
+      &:hover {
+        color: #00acdc;
```
TarasHoliuk, line 54: you should have used a variable here

src/styles/blocks/header/import_header.scss
TarasHoliuk: UPD: I see there are styles for the body in another file, so you probably forgot to delete it here

src/styles/blocks/main/main.scss
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+  font-family: Roboto, sans-serif;
+
+  .catalog {
```
TarasHoliuk, line 5: filename should match BEM block name
should be outside of `body` tag

TarasHoliuk (approved):
You've fixed all previous comments and generally GJ 👍 So approved ✔️ 

Added a few comments just FYI (no need to re-request review)

src/styles/blocks/main/main.scss
jooaasss: thanks for all sugestions, also i have a question about the nesting correctness in src/main/page.scss file. could you give me advise about that please? is it right way to do this?

jooaasss: I remember, that you said about re-request, but I'm interested in the thing above and would like to know the answer

PR #4352, opened 2024-05-07
add catalog solution

FaiHamid:
- [DEMO LINK](https://FaiHamid.github.io/layout_catalog/)
- [TEST REPORT LINK](https://FaiHamid.github.io/layout_catalog/report/html_report/)

- [x] Code follows all the [Code Style Rules ❗️]


FaiHamid:
Знімок екрана 2024-05-07 115122
I know that I again used * for box-sizing, but it didn`t work when I put box-sizing inside html or body tag. Should I add box-sizing straight into class when it need to be ?

dinaminahi (approved):
Well done ⭐

dinaminahi (commented):
Yes, better to set box-sizing to the element you need it on

PR #4370, opened 2024-05-09
Develop

ikocherovets:
- [DEMO LINK](https://ikocherovets.github.io/layout_catalog/)
- [TEST REPORT LINK](https://ikocherovets.github.io/layout_catalog/report/html_report/)
- [x] repeated sizes and special colors are put to variables


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,106 @@
...
+    background-color: $buttonBackgroundColor;
+    color: $buttonTextColor;
+
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    text-decoration: none;
+    
+    transition: 300ms;
```
roman-mirzoian, line 99: It's better to specify to what properties you need to apply the transition.

image

roman-mirzoian (changes requested):
Well done, but now your demo looks like this. Let's fix it.

image

src/styles/blocks/card.scss
ikocherovets: got it, thanks for these suggestions. Need to pay a more attention with the documentation.

src/styles/blocks/variables.scss
```diff
@@ -0,0 +1,21 @@
...
+$priceNumberColor: #060b35;
+$buttonHeight: 40px;
+$buttonBorderColor: #00acdc;
+$buttonBorderRadius: 5px;
+$buttonBackgroundColor: #00acdc;
+$buttonTextColor: #fff;
+$buttonHoverBackgroundColor: #fff;
+$buttonHoverTextColor: #00acdc;
+$buttonIsActive: #00acdc;
+$columnCount: 1;
```
dinaminahi, line 21: There's no need for separate variables if the same value is used multiple times. Variables are used to store values that might change, so if the same value is used in multiple places, it's more efficient to use the same variable instead of creating duplicates.  Let's reuse the same variable throughout your files for consistency.
Example: 

```
// Using a variable
$primary-color: #3498db;

.button {
  background-color: $primary-color;     // we use variable here
  color: darken($primary-color, 10%);  // and same variable here
}
```

dinaminahi, line 21:
```suggestion
$bodyFont: 'Roboto', sans-serif;
$cardWidth: 200px;
$borderColor: #f3f3f3;
$borderRadius: 5px;
$imageWidth: 160px;
$imageHeight: 134px;
$titleColor: #060b35;
$descriptionColor: #616070;
$starWidth: 16px;
$buttonHeight: 40px;
$buttonBackgroundColor: #00acdc;
$buttonTextColor: #fff;
$columnCount: 1;
```

dinaminahi (changes requested):
Good job 👍  But there is one thing we need to improve to complete the task

src/styles/blocks/variables.scss
ikocherovets: thanks a lot for this commentar. I thought that we can make some changes in the future with colors of button for example, so I created a lot of variables. Good to know.

dinaminahi (approved):
Yes 👍  Well done ⭐

PR #4414, opened 2024-05-29
Catalog solution Karaniewski

Karanelus:
- [DEMO LINK](https://Karanelus.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Karanelus.github.io/layout_catalog/report/html_report/)

[X] All component follow BEM and use SCSS
[X] repaeted sizes and special colors are put to variables
[X] Grid is used for the columns
[X] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
[X] All changes on `:hover` are smooth
[X] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,6 +22,360 @@
   </head>
 
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a
+        href="index.html"
+        class="header__logo"
+      >
+        <img
+          src="./images/logo.png"
+          alt="Moyo logo"
+        />
+      </a>
+
+      <nav class="nav header__nav">
+        <ul class="links nav__links">
+          <li class="link links__link">
+            <a
+              href="#Apple"
+              class="link__page is-active"
+            >
+              Apple
+            </a>
+          </li>
+
+          <li class="link links__link">
+            <a
+              class="link__page"
+              href="#Samsung"
+            >
+              Samsung
+            </a>
+          </li>
+
+          <li class="link links__link">
+            <a
+              class="link__page"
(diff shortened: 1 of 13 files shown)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,61 @@
...
+  padding: 0 50px;
+  box-sizing: border-box;
+  width: 100dvw;
+
+  &__logo {
+    width: 40px;
+    height: 40px;
+  }
+
+  .nav {
```
nataliaklonowska, line 16: nav is a separate block, it should be in separate file

src/index.html
```diff
@@ -22,6 +22,360 @@
...
+        class="header__logo"
+      >
+        <img
+          src="./images/logo.png"
+          alt="Moyo logo"
+        />
+      </a>
+
+      <nav class="nav header__nav">
+        <ul class="links nav__links">
```
nataliaklonowska, line 37: do you need links as separate BEM block?

nataliaklonowska (changes requested):
Great work 🔥 
Just evaluate if in header you need so many BEM blocks and move them to separate files :)

src/index.html
Karanelus: So i must remain only nav_links class? 
Because in css even list child (li > a) started from link

nataliaklonowska: you didn't use or needed class "links" but block link you could left

nataliaklonowska (approved):

PR #4465, opened 2024-06-21
Catalog

Kateryna99:
- [DEMO LINK](https://Kateryna99.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Kateryna99.github.io/layout_catalog/report/html_report/)

- [x ] All component follow BEM and use SCSS
- [x ] repaeted sizes and special colors are put to variables
- [x ] Grid is used for the columns
- [x ] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
- [ x] All changes on `:hover` are smooth
- [x] Code follows all the [Code Style Rules ❗️]


src/styles/stars.scss
```diff
@@ -0,0 +1,22 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    margin-right: 4px;
+  }
+
+  &--1 .stars__star:nth-child(-n + 1),
+  &--2 .stars__star:nth-child(-n + 2),
+  &--3 .stars__star:nth-child(-n + 3),
+  &--4 .stars__star:nth-child(-n + 4),
+  &--5 .stars__star:nth-child(-n + 5) {
```
DanilWeda, line 19: Try to use scss loop in this case.

DanilWeda (changes requested):
Great job! Perfect. 
Please describe to me why we need CSS files in the src folder.
Because now I do not understand the duplication of style files.
If we can not use it, remove it. In this case, we make the src folder more clear for review.
Also, check please the minor comment that I left. 
Many thanks)

src/styles/stars.scss
Kateryna99: Hah, typical 'copy-paste' problem :))

My bad, I missed Parcel is required for all  projects from now on. I will add it !

mrkovaliv (approved):
👍 👍 👍

PR #4471, opened 2024-06-24
add solution

coldshein:
- [DEMO LINK](https://coldshein.github.io/layout_catalog/)
- [TEST REPORT LINK](https://coldshein.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,6 +22,404 @@
   </head>
 
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a
+        href="#"
+        class="header__link"
+      >
+        <img
+          src="/src/images/logo.png"
+          alt="logo"
+          class="logo"
+        />
+      </a>
+      <nav class="header__nav">
+        <ul class="header__nav-list">
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link is-active"
+            >
+              apple
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              samsung
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              smartphones
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+              data-qa="nav-hover"
+            >
+              laptops & computers
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              gadgets
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              tablets
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              photo
+            </a>
+          </li>
+          <li class="header__nav-item">
+            <a
+              href="#"
+              class="header__nav-link"
+            >
+              video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="catalog">
+      <div
+        class="card"
+        data-qa="card"
+      >
+        <div class="card__img">
+          <img
+            src="./images/imac.jpeg"
+            alt="imac_logo"
+          />
+        </div>
+        <div class="card__container">
+          <h2 class="card__title">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h2>
+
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -22,6 +22,404 @@
...
+        href="#"
+        class="header__link"
+      >
+        <img
+          src="/src/images/logo.png"
+          alt="logo"
+          class="logo"
+        />
+      </a>
+      <nav class="header__nav">
```
ArturErmolenko, line 36: Using the header prefix, you bind the navigation block to this header.
BEM idea that you can reuse blocks.
For example, let’s just imagine that you have exactly the same navigation in your footer, but different footer styles
In this case, you will have to duplicate the code and write a separate navigation for the footer
Therefore, it is better to separate the navigation from the header so that it is a separate block
```suggestion
      <nav class="nav">
```

src/index.html
```diff
@@ -22,6 +22,404 @@
...
+        <ul class="header__nav-list">
```
ArturErmolenko, line 37:
```suggestion
        <ul class="nav-list">
```
And so on ))

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,83 @@
...
+      height: 100%;
+    }
+  }
+
+  &__nav {
+    display: block;
+    align-items: center;
+
+    &-list {
+      display: flex;
```
ArturErmolenko, line 27: So if you use separate block for nav, there the should be separated file for navigation

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,148 @@
...
+    background-repeat: no-repeat;
+    background-position: left center;
+  }
+
+  @for $star from 1 through 5 {
+    &--#{$star} :nth-child(-n + #{$star}) {
+      background-image: url(../images/star-active.svg);
+    }
+  }
+}
```
ArturErmolenko, line 148: Block stars should have separeted file

src/index.html
```diff
@@ -22,6 +22,404 @@
...
+        </div>
+        <div class="card__container">
+          <h2 class="card__title">
+            APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+          </h2>
+
+          <p class="card__code card-text">Product code: 195434</p>
+
+          <div class="card__rating">
+            <ul class="card__stars stars stars--4">
```
ArturErmolenko, line 399:
```suggestion
            <ul class="stars stars--4">
```
Do you need card_stars class, let stars block be independent so that we follow BEM

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,148 @@
+.card {
+  display: flex;
+  flex-direction: column;
+
+  width: $card-width;
+  height: 408px;
```
ArturErmolenko, line 6: Don't fix container size (if there is no such a requirement)
Let the content size dictate it. To avoid overflow or accidental scroll bar
```suggestion
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,148 @@
...
+    color: #fff;
+    text-transform: uppercase;
+    text-decoration: none;
+
+    background-color: $primary-color;
+
+    border-radius: 5px;
+    border: 1px solid $primary-color;
+
+    height: 40px;
```
ArturErmolenko, line 104: Same here
Don't fix container size (if there is no such a requirement)
Let the content size dictate it. To avoid overflow or accidental scroll bar
Use padding instead
```suggestion
```

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,148 @@
...
+
+    transition: all 0.3s ease-in-out;
```
ArturErmolenko, line 106:
```suggestion
    transition: background-color, color 0.3s ease-in-out;
```
Don't use the 'all' keyword in CSS transitions, it's bad for performance
[Here](https://www.pno.dev/articles/dont-use-the-all-keyword-in-css-transitions/) you can read more about it

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,83 @@
...
+  height: 60px;
+
+  font-family: Roboto, sans-serif;
+  font-size: 12px;
+  font-weight: 500;
+  text-decoration: none;
+  text-transform: uppercase;
+
+  color: $main-color;
+  transition: all 0.2s;
```
ArturErmolenko, line 60: Try not to use all )

ArturErmolenko (changes requested):
Nice start 🔥 , let's improve your code by fixing comments )
And one more I see from your Demo link that you change button color on card hover, it shouldn't work this way, button color should change only when I hover the button))

VenomMB (approved):
Good Job!!!

PR #4493, opened 2024-06-29
Layout_catalog

igor-romasiuk:
- [DEMO LINK](https://igor-romasiuk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://igor-romasiuk.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,101 @@
+.card {
+  width: 200px;
+  background-color: #fff;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  transition: transform 300ms;
+
+  &:hover {
+    transform: scale(1.2);
+  }
+
+  &__photo {
+    display: block;
+    width: 160px;
+    height: 134px;
+    margin: 0 auto 40px;
+  }
+
+  &__title {
+    margin: 0;
+    margin-bottom: 0;
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 18px;
+    color: #060b35;
+
+    &:hover {
+      color: #34568b;
+    }
+  }
+
+  &__code {
+    font-weight: 400;
+    line-height: 14px;
+    margin: 4px 0 16px;
+    color: #616070;
+  }
+
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: flex-end;
+
+    &:not(:last-child) {
+      margin-bottom: 24px;
+    }
+  }
+
+  &__review-count {
+    margin: 0;
+    line-height: 16px;
+    color: #060b35;
+  }
+
+  &__price {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 16px;
+  }
+
+  &__price-label {
+    font-size: 12px;
+    line-height: 18px;
+    color: #616070;
+  }
+
+  &__price-value {
+    font-size: 16px;
+    line-height: 18px;
+    font-weight: 700;
+    color: #060b35;
+  }
+
+  &__buy-button {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    box-sizing: border-box;
+    height: 40px;
+    border: 1px solid #00acdc;
+    border-radius: 5px;
+
+    font-size: 14px;
+    line-height: 16px;
+    color: #fff;
+    font-weight: 700;
+    text-decoration: none;
+    text-transform: uppercase;
+
+    background-color: #00acdc;
+  }
+
+  &__buy-button:hover {
+    color: #00acdc;
+    background-color: #fff;
+  }
+}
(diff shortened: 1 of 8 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+    &:hover {
+      color: #34568b;
+    }
+  }
+
+  &__code {
+    font-weight: 400;
+    line-height: 14px;
+    margin: 4px 0 16px;
+    color: #616070;
```
BudnikOleksii, line 38: Create variables for colors and transition duration as well so it will be the same for all elements

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,28 @@
...
+@media (min-width: 768px) {
+  .catalog {
+    display: grid;
+    grid-template-columns: repeat(3, 200px);
+  }
+}
+
+@media (min-width: 1024px) {
+  .catalog {
+    display: grid;
```
BudnikOleksii, line 25: Use nesting

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,59 @@
...
+    width: 40px;
+  }
+
+  .nav__list-items {
+    display: flex;
+    list-style: none;
+    margin: 0;
+    padding: 0;
+
+    .nav__item {
```
BudnikOleksii, line 19: - Use nesting
- Create a separate file per each styles block
image

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,20 @@
...
+  display: flex;
+  width: 16px;
+  height: 16px;
+  background-image: url(/src/images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+@for $i from 0 through 5 {
+  .stars--#{$i} .stars__star:nth-child(-n + #{$i}) {
```
BudnikOleksii, line 17: Nesting

src/styles/index.scss
```diff
@@ -1,3 +1,14 @@
+html {
+  font-family: Roboto, sans-serif;
+  font-size: 10px;
+}
+
 body {
   margin: 0;
+  box-sizing: border-box;
 }
```
BudnikOleksii, line 9: Move to reset/normalize file

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution.
Also, "Make all the changes smooth on hover (during 300ms)", and when you hover button add the cursor pointer, this a good practice so a knows that they can interact with the element

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
...
+    text-decoration: none;
+    text-transform: uppercase;
+
+    background-color: $white--blue--color;
+  }
+
+  &__buy-button:hover {
+    color: $white--blue--color;
+    background-color: $white--color;
+  }
```
BudnikOleksii, line 100: Use nesting, and the button still don't have any transitions

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,101 @@
+.card {
+  width: 200px;
+  background-color: $white--color;
+  border: 1px solid $white--grey--color;
+  border-radius: 5px;
+  padding: 32px 16px 16px;
+  box-sizing: border-box;
+  transition: $transition;
```
BudnikOleksii, line 8: Specify to which properties should transition apply

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,28 @@
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(1, 200px);
+  justify-content: center;
+  gap: 46px 48px;
+  padding: 50px 40px;
+
+  @media (min-width: 488px) {
+    & {
+      display: grid;
```
BudnikOleksii, line 10: - You don't need to set grid in media queries as the `.catalog` selector itself has this style
- and it's not necessary to specify `& {`

src/styles/blocks/navigation.scss
```diff
@@ -0,0 +1,46 @@
...
+    display: flex;
+    position: relative;
+    margin-left: 20px;
+    height: 60px;
+
+    &:first-child {
+      margin-left: 0;
+    }
+
+    .nav__link-item {
```
BudnikOleksii, line 17: You don't need to nest selectors like this as in this case, you have a selector like this:
image

They all just should be inside `nav`
```scss
.nav {
  &__list {}

  &__item {}
  // other
}
```

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,7 @@
+$white--blue--color: #00acdc;
+$dark--blue--color: #060b35;
+$white--color: #fff;
+$white--grey--color: #f3f3f3;
+$dark--grey--color: #616070;
+$blue--color: #34568b;
```
BudnikOleksii, line 6: Variable names should show their purpose. If the color palette changes and all elements that uses blue color will use red instead you have to find and rename in all places

BudnikOleksii (changes requested):
Almost done, just a few moments should be fixed

mrkovaliv (approved):
👍

PR #4499, opened 2024-07-01
add solution

PolianskyiDmytro:
- [DEMO LINK](https://PolianskyiDmytro.github.io/layout_catalog/)
- [TEST REPORT LINK](https://PolianskyiDmytro.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,0 +1,56 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+}
+
+.logo {
+  align-self: center;
+  margin-left: 50px;
+}
+
+.nav {
+  display: flex;
+  height: 60px;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,56 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+}
+
+.logo {
```
TarasHoliuk, line 6: I understand that the header and logo are small BEM blocks
But according to BEM rules, each BEM block should be in a separate file

So they should be in separate files (or make the logo just an element of the header block)

Not critical and everything else looks good so approved ✔️ 
But remember this in the future 🙂

TarasHoliuk (approved):
GJ 👍 
1 comment just FYI

PR #4522, opened 2024-07-14
add task solution

Syogunok:
- [DEMO LINK](https://Syogunok.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Syogunok.github.io/layout_catalog/report/html_report/)


TarasHoliuk:
According to test results there are things that should be fixed:
image

PAdding top for the main tag:
image

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+$pad-box: 16px;
+$size-img: calc($size-box - $size-box / 5);
+$size-40: 40px;
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
 }
+
+.header {
```
TarasHoliuk, line 12: Each BEM block should be in a separate file according to BEM-rules

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+    background-color: #{$color-link};
+  }
+
+  &__link.is-active,
+  &__link:hover {
+    color: #{$color-link};
+  }
+}
+
+ul {
```
TarasHoliuk, line 80: you can just add this styles using nav__list selector

same for `li` below

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+
+@media (min-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(4, 200px);
+  }
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
TarasHoliuk, line 124: default value - no need to set it

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+  gap: 7.2px;
+  margin-left: 1.3px;
+
+  &__star {
+    width: 12.8px;
```
TarasHoliuk, line 129: strange size
Should be 16px

image

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+  .catalog {
+    grid-template-columns: repeat(4, 200px);
+  }
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  gap: 7.2px;
+  margin-left: 1.3px;
```
TarasHoliuk, line 126: just use the correct margin (4px)

Note that the margin should be only between stars (like gap, so basically you can use gap, but make sure you set correct value and tests pass), so no left margin for first star or vice versa - no right margin for the last star

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+    margin-top: 18px;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+
+    &-text {
+      color: #060b35;
+      font-size: 10px;
+      line-height: 14px;
+      font-weight: 400;
```
TarasHoliuk, line 182: it's default value - no need to set it

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+    font-size: 10px;
+    line-height: 14px;
+    font-weight: 400;
+    margin-top: 4px;
+  }
+
+  &__review {
+    margin-top: 18px;
+    display: flex;
+    flex-direction: row;
```
TarasHoliuk, line 175: it's default value - no need to set it

fix all similar cases

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+
+    &-res {
+      font-size: 16px;
+      line-height: 18px;
+      font-weight: 700;
+    }
+  }
+}
+
+.but {
```
TarasHoliuk, line 206: what is but?
do not use short forms

src/styles/index.scss
```diff
@@ -1,3 +1,243 @@
...
+    &-text {
+      color: #060b35;
+      font-size: 10px;
+      line-height: 14px;
+      font-weight: 400;
+    }
+  }
+
+  &__price {
+    margin-top: 24px;
```
TarasHoliuk, line 187: make sure it has correct margin (or elements above) - tests fail

TarasHoliuk (changes requested):

src/styles/index.scss
```diff
@@ -1,3 +1,44 @@
...
+@import 'header';
+@import 'catalog.place';
+@import 'card';
+
 body {
   margin: 0;
+  font-family: Roboto, sans-serif;
 }
+
+.button {
```
TarasHoliuk, line 12: let's also move it to a separate file (or make it an element of card) - it's a separate BEM block

src/styles/header.scss
```diff
@@ -0,0 +1,80 @@
...
+    background-color: #{$color-link};
+  }
+
+  &__link.is-active,
+  &__link:hover {
+    color: #{$color-link};
+  }
+}
+
+ul {
```
TarasHoliuk, line 69: not fixed
image

Actually, it seems redundant at all (almost all styles are overwritten in nav__list)

src/styles/header.scss
```diff
@@ -0,0 +1,80 @@
...
+  list-style-type: disc;
+  margin-block: 1em;
+  margin-inline: 0;
+  padding-inline-start: #{$size-40};
+}
+
+li {
+  display: list-item;
+  text-align: -webkit-match-parent;
+}
```
TarasHoliuk, line 80: remove it

TarasHoliuk (changes requested):
Almost done 👍 
A few more fixes are required

If you have any question feel free to ask them in the chat

lerastarynets (approved):
lgtm 🚀

PR #4559, opened 2024-08-09
Added layout

CamilleHaus:
Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
- [DEMO LINK](https://CamilleHaus.github.io/layout_catalog/)
- [TEST REPORT LINK](https://CamilleHaus.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -22,6 +22,327 @@
...
+      <div
+        class="card__container"
+        data-qa="card"
+      >
+        <img
+          src="./images/imac.jpeg"
+          alt="Imac A1419"
+          class="card__container-image"
+        />
+        <h2 class="card__container-title">
```
pedro-ruas, line 116: Camille, there isn't nothing wrong with this name, ok? But it could be simplified a little.
What you are saying is: "Container Title" is an _Element_ of the _Block_ "Card". And this is ok.
But I believe you mistakenly thought that you needed to add "__container" to every element's name, what isn't the case.
The name "card__title" would work just as fine, and this is valid for all following elements ;D

src/index.html
```diff
@@ -22,6 +22,327 @@
...
+              class="nav__list-link"
+              href="#"
+            >
+              Video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <main class="card">
```
pedro-ruas, line 106: "card-list" or "card-grid" or just "grid" would all be better names here, because this doesn't refer to a single card, but the whole group of it

pedro-ruas (changes requested):
Camille, your code looks well made, and there are minimal improvements that could be done.
However, I see your DEMO link isn't working, it leads to a 404 page, so I cannot properly test it.
Another thing is that your TEST link works, but it says your code failed some tests and thus I cannot approve it until you pass the automated tests.
Test everything in your machine and when all tests pass, send it again for review ;D

PR #4571, opened 2024-08-14
add solution

Tulenty:
- [DEMO LINK](https://Tulenty.github.io/layout_catalog/)
- [TEST REPORT LINK](https://Tulenty.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,6 +22,388 @@
   </head>
 
   <body>
-    <h1>Catalog</h1>
+    <header class="header">
+      <a
+        href="#"
+        class="logo"
+      >
+        <img
+          src="./images/logo.png"
+          alt="Moyo logo"
+          class="logo__image"
+        />
+      </a>
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link is-active"
+            >
+              Apple
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+            >
+              Samsung
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              href="#"
+              class="nav__link"
+            >
+              Smartphones
+            </a>
+          </li>
+          <li class="nav__item">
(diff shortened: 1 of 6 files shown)


src/styles/card.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  &__image {
+    display: block;
+    margin: 0 auto 40px;
+    width: 160px;
+    height: 134px;
+  }
+
+  &__title {
+    margin: 0 0 4px;
```
TarasHoliuk, line 16:
```suggestion
    margin-bottom: 4px;
```

src/styles/index.scss
```diff
@@ -1,3 +1,50 @@
...
+
 body {
   margin: 0;
 }
+
+p {
+  margin: 0;
+}
+
+.catalog {
```
TarasHoliuk, line 24: even if it's small it is a separate BEM block, so should be in a separate file according to BEM rules

src/styles/header.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.nav__list {
+  display: flex;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+}
+
+.nav__item {
```
TarasHoliuk, line 26: just FYI, you can also do it in this way:

```suggestion
.nav__item:not(:first-child) {
```

src/index.html
```diff
@@ -22,6 +22,388 @@
...
+          alt="iMac A1419"
+        />
+
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+
+        <p class="card__code">Product code: 195434</p>
+
+        <section class="card__review-container">
```
TarasHoliuk, line 346: the section is usually something bigger (that has its own title). It's not a mistake here and you can keep the section, just remember that div is also allowed to use 🙂

TarasHoliuk (approved):
Great job!

I added a few comments just FYI. Approved ✔️

PR #4581, opened 2024-08-16
Develop

hma-3:
- [DEMO LINK](https://hma-3.github.io/layout_catalog/)
- [TEST REPORT LINK](https://hma-3.github.io/layout_catalog/report/html_report/)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,11 @@
+$main-accent-color: #060b35;
+$secondary-color: #616070;
+$blue-accent-color: #00acdc;
+$card-border-color: #f3f3f3;
+$header-bg-color: #fff;
+$header-shadow-color: #0000000d;
+$title-hover-color: #34568b;
+$card-color: #fff;
```
hma-3, line 8: I have two white colors in case we will want to change color only for header, or only for card

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,11 @@
...
+$header-logo-size: 40px;
```
hma-3, line 9: this size is used only for width and height of logo image, but it should be always a square, so I added a variable

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,62 @@
...
+.logo {
+  display: flex;
+
+  &__img {
+    width: $header-logo-size;
+    height: $header-logo-size;
+  }
+}
+
+.nav {
```
BudnikOleksii, line 22: Create a separate file for each BEM block

BudnikOleksii (approved):
Good work 👍

PR #4590, opened 2024-08-18
add task solution

WhyCame:
Look at BackstopJS Report, I don't understand what the problem is, maybe in a test? Based on the fact that one test shows that there is a problem with objects inside the product map, but on the product map check everything is ok. How can this be?

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,54 @@
+@import 'variables';
+@import 'product-card';
+@import 'header';
+
+.showcase {
+  --column-count: 4;
+
+  display: grid;
+
+  grid-template-columns: repeat(var(--column-count), 200px);
+  padding: 50px 40px;
+  justify-content: center;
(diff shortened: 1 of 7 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.showcase {
+  --column-count: 4;
+
+  display: grid;
+
+  grid-template-columns: repeat(var(--column-count), 200px);
+  padding: 50px 40px;
+  justify-content: center;
+  gap: 48px 46px;
```
vadiimvooo, line 13: try to switch this params to pass tests.
```suggestion
   gap: 46px 48px;
```

vadiimvooo (changes requested):
Please try to pass all tests. Feel free to ask for help in fe_chat.

SanyaBratashchuk (approved):
LGTM

PR #4598, opened 2024-08-19
add catalog

Valeraom:
[DEMO LINK](https://Valeraom.github.io/layout_catalog/)
[TEST REPORT LINK](https://Valeraom.github.io/layout_catalog/report/html_report/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,119 @@
...
+    color 300ms linear;
+
+  &:hover {
+    cursor: pointer;
+    transform: scale(120%);
+  }
+
+  &__title {
+    display: block;
+    margin: 0 0 4px;
```
VolodymyrKirichenko, line 24: `margin-bottom`?

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,8 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+
+  box-shadow: 0 2px 4px 0 #0000000d;
```
VolodymyrKirichenko, line 7: use variables for all colors

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,8 @@
+.page {
+  font-family: Roboto, Arial, sans-serif;
+  font-weight: 400;
+
+  &__body {
+    margin: 0;
+  }
+}
```
VolodymyrKirichenko, line 8: You can use `html` or `body` tag for it and rename this file. It will now be `reset.scss` instead of `page.scss`.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,20 @@
+$logo-width: 40px;
+$logo-height: 40px;
+$card-width: 200px;
+$card-photo-width: 160px;
+$card-photo-height: 134px;
+$card-star-width: 16px;
+$card-star-height: 16px;
```
VolodymyrKirichenko, line 7: but why? u only used it once

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,20 @@
...
+$nav-link-color: #060b35;
+$nav-color-hovered: #00acdc;
+$nav-color-activated: #00acdc;
+$nav-line-bg-color-activated: #00acdc;
```
VolodymyrKirichenko, line 11: Do not duplicate variables for the same colors(In all the cases!)

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  &__rating {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 24px;
+  }
+
+  &__stars {
```
VolodymyrKirichenko, line 63: but why?
u already have a separate `stars.scss` file for stars and `stars` class for it
use it. Don't create a new one

src/index.html
```diff
@@ -1,27 +1,446 @@
...
+
+        <div class="card__price-container">
+          <span class="card__price-label">Price:</span>
+
+          <span class="card__price-value">$2,199</span>
+        </div>
+
+        <a
+          href="#"
+          class="card__buy-link"
```
VolodymyrKirichenko, line 322: it's not a link, it's a button)
```suggestion
          class="card__buy-button"
```

src/index.html
```diff
@@ -1,27 +1,446 @@
...
     />
   </head>
 
-  <body>
-    <h1>Catalog</h1>
+  <body class="page__body">
+    <header class="header">
+      <div class="logo">
+        <a
+          href="https://"
```
VolodymyrKirichenko, line 36: ur link is broken.
fix it
```suggestion
          href="#"
```

VolodymyrKirichenko (changes requested):
Your code is noticeably better, but let's make a few corrections and fix a couple of mistakes 🥹

src/styles/blocks/card.scss
Valeraom: @VolodymyrKirichenko I reset margin-top in this line too

src/styles/blocks/card.scss
Valeraom: @VolodymyrKirichenko I did it to avoid setting position properties for block, as BEM says
Or if needed to reset I can do it?

src/index.html
Valeraom: @VolodymyrKirichenko I used "card__buy-button" before
but Kyrylo said, if we use tag `<a></a>` we need to name class with "link" and I renamed it

src/styles/utils/variables.scss
Valeraom: @VolodymyrKirichenko if we have the same color, but for different purposes, we also need to use one variable?
What to do, if I will need to set different colors for activated `.nav__link` and `.card__button`, for example?

src/styles/blocks/card.scss
VolodymyrKirichenko: @Valeraom you use the stars block according to the BEM.
This is a separate block that can be used in any part of the code.
additional class `card__stars` is redundant

src/styles/utils/variables.scss
VolodymyrKirichenko: > @VolodymyrKirichenko if we have the same color, but for different purposes, we also need to use one variable? What to do, if I will need to set different colors for activated `.nav__link` and `.card__button`, for example?

@Valeraom u can use next variant for color variables or something like that
u will always have a palette of colors used in the project and u can add it at the beginning

Знімок екрана 2024-08-19 о 21 32 32

src/index.html
VolodymyrKirichenko: > @VolodymyrKirichenko I used "card__buy-button" before but Kyrylo said, if we use tag `<a></a>` we need to name the class with "link" and I renamed it

@Valeraom  
`card__buy-link` - how to understand that this is a button? If I hadn't seen the demo, I would have thought it was just a link

Use `card__button` to explicitly indicate that the `<a>` tag functions as a button, even if it has a different tag.
it will be more clear. You'll always know it's a button.

Valeraom:
I didn't delete `$card-width` because I use it in mixins.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,8 @@
+$card-width: 200px;
+$midnight: #060b35;
+$black-shadow: #0000000d;
+$blue: #00acdc;
+$grey-blue: #616070;
+$white: #fff;
+$white-smoke: #f3f3f3;
+$dark-blue: #34568b;
```
BudnikOleksii, line 8: Variable name should describe the purpose as if the color in the palette changes to red for example you have to change the name in all places where you use it

BudnikOleksii (approved):
Good work 👍

src/index.html
Valeraom: @VolodymyrKirichenko ok, probably I misunderstood this, because Kyrylo used your variant in today's meeting)

PR #4601, opened 2024-08-19
Develop

coldmelnyk:
- [DEMO LINK](https://coldmelnyk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://coldmelnyk.github.io/layout_catalog/report/html_report/)

- [X] All component follow BEM and use SCSS
- [X] repaeted sizes and special colors are put to variables
- [X] cards are shown in 1, 2, 3 or 4 columns based on screen resolution


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  @media (min-width: 1024px) {
+    & {
+      grid-template-columns: repeat(4, 200px);
+    }
+  }
+
+  &__card {
```
BudnikOleksii, line 26: Card should be a separate BEM block as you can reuse it in other places

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,134 @@
...
+    width: 200px;
+    padding: 32px 16px 16px;
+    display: flex;
+    justify-content: space-between;
+    flex-direction: column;
+
+    background-color: $white-color;
+    border: 1px solid $border-color;
+    border-radius: 5px;
+    transition-duration: 300ms;
```
BudnikOleksii, line 36: Specify to which properties should transition apply

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,68 @@
...
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.05);
+
+  &__icon {
+    display: flex;
+    height: 40px;
+    width: 40px;
+    margin-left: 50px;
+  }
+
+  &__nav {
```
BudnikOleksii, line 19: The same, `nav` is a separate BEM block

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+}
+
+.is-active {
+  color: $blue-color;
+  position: relative;
+}
+
+.is-active::after {
```
BudnikOleksii, line 59: Use nesting

src/styles/utils/resets-normalize.scss
```diff
@@ -0,0 +1,7 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+
+  font-family: Roboto, Arial, Helvetica, sans-serif;
```
BudnikOleksii, line 6: - Don't set margins/paddings to all elements, reset only for tags that actually need it
- The same regarding font-family, instead you can specify it to html or body tag

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,9 @@
+$info-color: #616070;
+$review-color: #060b35;
+$border-color: #f3f3f3;
+$white-color: #fff;
+$star-size: 16px;
+$background-color-page: grey;
+$blue-color: #00acdc;
```
BudnikOleksii, line 7: Variable name should describe the purpose, as if this color in palette changes to red for example you have to change the name in all places where you ue it

BudnikOleksii (changes requested):
Good work 👍 
Left some comments and suggestions

src/styles/utils/resets-normalize.scss
coldmelnyk: > * Don't set margins/paddings to all elements, reset only for tags that actually need it
> * The same regarding font-family, instead you can specify it to html or body tag

I've moved font-family from `*` to the `html `tag, but I wanna do full reset of margins and padding because I knew that I'm gonna set it manually through all my work. So that's the reason - I don't wanna get errors with test because of some random padding of li or ul or some other elements.

BudnikOleksii: > > * Don't set margins/paddings to all elements, reset only for tags that actually need it
> > * The same regarding font-family, instead you can specify it to html or body tag
> 
> I've moved font-family from `*` to the `html `tag, but I wanna do full reset of margins and padding because I knew that I'm gonna set it manually through all my work. So that's the reason - I don't wanna get errors with test because of some random padding of li or ul or some other elements.

Ok, but it's better to reset only for tags that actually have them

BudnikOleksii (approved):
Looks good 🔥

PR #4663, opened 2024-09-09
add task solution

MrZefirchk:
I couldn't fix deploy, it doesn't work
image

- [DEMO LINK](https://MrZefirchk.github.io/layout_catalog/)
- [TEST REPORT LINK](https://MrZefirchk.github.io/layout_catalog/report/html_report/)

- [ ] All component follow BEM and use SCSS
- [ ] repaeted sizes and special colors are put to variables
- [ ] Grid is used for the columns
- [ ] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
- [ ] All changes on `:hover` are smooth
- [ ] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/styles/blocks/card.scss
+++ b/src/styles/blocks/card.scss
@@ -1,0 +1,107 @@
+.card {
+  width: $card-size;
+  padding: 32px 16px 16px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  box-sizing: border-box;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+
+  @include hover(
+    (
+      transform: scale(1.2),
+    )
+  );
+
+  &__container {
+    display: flex;
+    justify-content: center;
+    margin-bottom: 40px;
+  }
+
+  &__image {
+    width: 160px;
+    height: 134px;
+    background-position: center;
+  }
+
+  &__title {
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $main-text-color;
+    margin: 0 0 4px;
+  }
+
+  &:hover &__image {
+    color: #34568b;
+  }
+
+  &__product-code {
+    font-weight: 400;
+    font-size: 10px;
+    line-height: 14px;
+    color: $secondary-text-color;
+    margin: 0 0 16px;
+  }
+
+  &__rating-reviews {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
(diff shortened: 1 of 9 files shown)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
...
+    height: 134px;
+    background-position: center;
+  }
+
+  &__title {
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $main-text-color;
+    margin: 0 0 4px;
```
TarasHoliuk, line 34:
```suggestion
    margin-bottom: 4px;
```
same in the product-code

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
...
+
+  &__rating-reviews {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 24px;
+  }
+
+  &__reviews {
+    font-weight: 400;
```
TarasHoliuk, line 57: 400 is the default font-weight (same as normal - check mdn page about font-weight), so no need to set it

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,107 @@
...
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 16px;
+
+    &-label {
+      font-weight: 400;
+      font-size: 12px;
+      line-height: 18px;
+      color: $secondary-text-color;
+      margin: 0;
```
TarasHoliuk, line 75: I guess you don't need default margin in all `p` tags, maybe in h2 or body tags. So you can create _reset.scss and reset these values

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,94 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+}
+
+.logo-link {
+  width: 40px;
+  height: 40px;
+  line-height: 60px;
+}
+
+.nav {
```
TarasHoliuk, line 17: It's not critical here, but theoretically `nav` is a separate BEM block, so it should be in a separate file according to BEM rules. Same for logo-link (or make it element of header block)

TarasHoliuk (changes requested):
It almost perfect - great job 🔥 
I added a few tiny comments only)

TarasHoliuk:
> I couldn't fix deploy, it doesn't work

I see it works now, GJ 👍 But in case you need help it's better to join Q&A session or ask for help in the chat 🙂

TarasHoliuk (approved):

PR #4665, opened 2024-09-11
Develop

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/layout_catalog/)
[TEST REPORT LINK](https://14lucas-mendes.github.io/layout_catalog/report/html_report/)


pedro-ruas (changes requested):
Lucas, there are failing tests:
- Spacing between the header and the catalog itself is different, solving that will probably pass a couple tests already
- I believe the color from the letters in the header is also wrong and that is causing the fail for the header tests, double check the checklist as I believe there this is mentioned
- Your header seems to be missing the bottom border
- You did not implement any of the hovering effects for the card, double-check the checklist about that too

pedro-ruas (changes requested):
Lucas, the height of each card is slightly wrong, they start at the correct point in the page, but the first row ends later than it should and then the second row starts later than it should.
Also, `reviews` text is not centered with the stars

pedro-ruas (approved):
Nice job!!!!

PR #4807, opened 2024-10-25
add task solution

ErasoumZe:
- [DEMO LINK](https://ErasoumZe.github.io/layout_catalog/)
- [TEST REPORT LINK](https://ErasoumZe.github.io/layout_catalog/report/html_report/)

- [ ] All component follow BEM and use SCSS
- [ ] repaeted sizes and special colors are put to variables
- [ ] Grid is used for the columns
- [ ] cards are shown in 1, 2, 3 or 4 columns based on screen resolution
- [ ] All changes on `:hover` are smooth
- [ ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)


src/index.html
```diff
@@ -21,7 +24,374 @@
...
+            <a
+              class="nav__link"
+              href="#"
+            >
+              Smartphones
+            </a>
+          </li>
+          <li class="nav__item">
+            <a
+              class="nav__link nav__link--hover"
```
pedro-ruas, line 68: This hover should work for all links. There is no need for a class `nav__link--hover`, you can use the hovering styling for `nav__link`

src/index.html
```diff
@@ -21,7 +24,374 @@
...
+          src="./images/imac.jpeg"
+          alt="imac"
+        />
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <span class="card__code">Product code: 195434</span>
+
+        <div class="card__rating">
+          <div class="stars stars--active">
```
pedro-ruas, line 127: Here you should use the same block from the Stars task. Meaning that stars should have a modifier for `stars--4` and the styles for it should work as is on that task.
You should not fix the styling for 4 stars

src/index.html
```diff
@@ -17,11 +20,378 @@
...
+        </div>
+
+        <div class="card__price">
+          <div class="card__price-text">Price:</div>
+          <div class="card__price-number">$2,199</div>
+        </div>
+
+        <a
+          href="#"
+          class="buy"
```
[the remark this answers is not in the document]
pedro-ruas: Ignore this AI comment, data-qa is necessary

src/index.html
```diff
@@ -21,7 +24,374 @@
...
+          src="./images/imac.jpeg"
+          alt="imac"
+        />
+        <h2 class="card__title">
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h2>
+        <span class="card__code">Product code: 195434</span>
+
+        <div class="card__rating">
+          <div class="stars stars--active">
```
pedro-ruas, line 163: Same comment as before is valid for all stars block in this task

src/index.html
```diff
@@ -21,7 +24,374 @@
...
+        </div>
+
+        <div class="card__price">
+          <div class="card__price-text">Price:</div>
+          <div class="card__price-number">$2,199</div>
+        </div>
+
+        <a
+          href="#"
+          class="buy"
```
pedro-ruas, line 390: There should be a better naming for this class, as this is not following BEM.
You can use something like `card__buy-button`

src/styles/blocks/buy.scss
```diff
@@ -0,0 +1,19 @@
...
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+  border-radius: $rounding;
+  transition: $transition-duration;
+
+  &:hover {
+    background-color: $white;
+    color: $blue-accent;
+    border: $blue-border;
```
pedro-ruas, line 17: When hovering the button, you can see that all blocks do a small "jump". The problem is that the button has no borders when not hovered, but gets borders added when you hover.
To fix this issue, you can pass this `border` property to the initial state of the class (without the hover).
Because the border is the same color of the background when not hovering, we won't notice it there, but doing so will prevent that "jump" from happening.
You could also use the same width border in the initial state, but with transparent color, if you wish

src/styles/blocks/nav.scss
```diff
@@ -0,0 +1,42 @@
...
+
+  &__link {
+    position: relative;
+    display: block;
+    text-decoration: none;
+    color: $main-accent;
+    padding-block: 23px;
+    transition: $transition-duration;
+
+    &--hover:hover {
```
pedro-ruas, line 22: Remove `--hover` and use this styling for `&:hover` and it will be perfect

src/styles/blocks/stars.scss
```diff
@@ -0,0 +1,19 @@
+.stars {
```
pedro-ruas, line 1: Copy the styling for that other task called Stars

pedro-ruas (changes requested):
Almost perfect, but needing a few changes. Great job!

ErasoumZe:
@pedro-ruas , fiz as modificações que pediu mas ainda não passo nos testes, e não consegui entender o porque... se puder de dar uma luz por favor, obrigado

julianFBarbosa:
Olá @ErasoumZe! Estava dando uma olhada na sua PR e parece que ela possui conflitos, pode resolvê-los por gentileza?

src/styles/blocks/buy.scss
```diff
@@ -0,0 +1,20 @@
+.card__buy-button {
```
raulriato, line 1: The size of the buttons are not corresponding to the referene because your trying to define the size of the button by just using padding. It won't work in this case. It's better to use the height and width properties to define the size.

src/index.html
```diff
@@ -21,7 +24,372 @@
...
+          Buy
+        </a>
+      </div>
+
+      <div class="card">
+        <img
+          class="card__image"
+          src="./images/imac.jpeg"
+          alt="imac"
+        />
```
raulriato, line 338: I believe you forgot to adjust this card's information haha

src/index.html
```diff
@@ -21,7 +24,372 @@
     />
   </head>
 
-  <body>
-    <h1>Catalog</h1>
+  <body class="page__body">
```
raulriato: Following the BEM methodology, your body is a block, not an element, so it's better to use only `page` as the class, instead of `page__body`

src/index.html
```diff
@@ -21,7 +24,372 @@
...
+          src="./images/logo.png"
+          alt="logo"
+        />
+      </a>
+
+      <nav class="nav">
+        <ul class="nav__list">
+          <li class="nav__item">
+            <a
+              class="nav__link nav__link--active is-active"
```
raulriato: You're already using `nav__link--active`. It's not good to use the `is-active` class together, as it can create redundancy in your code.

raulriato (changes requested):
You're almost there. Good job!

src/index.html
ErasoumZe: yes hahahha

joaorpereira (approved):

PR #4910, opened 2024-12-07
Develop

olga-wukelan:
- [DEMO LINK](https://olga-wukelan.github.io/layout_catalog/)
- [TEST REPORT LINK](https://olga-wukelan.github.io/layout_catalog/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -118,9 +118,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
@@ -155,9 +155,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
@@ -187,9 +187,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
@@ -219,9 +219,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
@@ -251,9 +251,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
@@ -283,9 +283,9 @@
           src="../src/images/imac.jpeg"
           alt="imac"
         />
-        <h1 class="card__title">
+        <h2 class="card__title">
           APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
-        </h1>
+        </h2>
         <p class="card__code">Product code: 195434</p>
         <!-- #region stars -->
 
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -17,11 +18,408 @@
...
+            >
+              video
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <!-- #endregion -->
+
+    <div class="main">
```
TarasHoliuk: it should be main tag (the reason part of the tests failed)

src/index.html
```diff
@@ -17,11 +18,408 @@
...
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+            <div class="stars__star"></div>
+          </div>
+          <h1 class="card__reviews">Reviews: 5</h1>
+        </div>
+
+        <div class="card__value">
+          <h1 class="card__price">Price:</h1>
```
TarasHoliuk: h1 is not a proper tag for all text - use `p` instead

h1 is a title of the entire page

src/index.html
```diff
@@ -17,11 +18,408 @@
...
+      <div
+        class="card"
+        data-qa="card"
+      >
+        <img
+          class="card__image"
+          src="../src/images/imac.jpeg"
+          alt="imac"
+        />
+        <h1 class="card__title">
```
TarasHoliuk:
```suggestion
        <h2 class="card__title">
```
h1 is a title of the entire page

src/index.html
```diff
@@ -17,11 +18,408 @@
...
+          APPLE A1419 iMac 27" Retina 5K Monoblock (MNED2UA/A)
+        </h1>
+        <h1 class="card__code">Product code: 195434</h1>
+
+        <div class="card__active">
```
TarasHoliuk: why `active`? you can use smth like `reviews-block`, etc.

src/index.html
```diff
@@ -17,11 +18,408 @@
...
+        >
+          Buy
+        </button>
+      </div>
+      <!-- #endregion -->
+
+      <!-- #region card -->
+      <div
+        class="card"
+        data-qa="card"
```
TarasHoliuk: you should add data-qa attributes only to the first card (to avoid tests duplication) - check the task description

src/index.html
```diff
@@ -17,11 +18,408 @@
...
+      >
+        <img
+          class="card__image"
+          src="../src/images/imac.jpeg"
```
TarasHoliuk, line 187:
```suggestion
          src="./images/imac.jpeg"
```
should work

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  &__code {
+    font-family: $main-font-family;
+    font-size: 10px;
+    font-weight: $normal;
+    line-height: 14px;
+    color: $secondary-color;
+    margin-top: 4px;
+    margin-bottom: 16px;
```
TarasHoliuk: it's better to set only top o only bottom (read about margin collapse)
You've already set margin-bottom for title - no need to set margin-top: 4px here:

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
+.card {
```
TarasHoliuk, line 1: fix size of the card: 

image

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
...
+  &__active {
+    display: flex;
+    justify-content: space-between;
+    height: 16px;
+    margin-bottom: 24px;
+    align-items: center;
+  }
+
+  &__reviews {
+    font-family: $main-font-family;
```
TarasHoliuk: it will be inherited- set it on body level tag and remove everywhere else

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
...
+    font-weight: $normal;
```
TarasHoliuk: no need to set it - it's default value

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
...
+    margin-bottom: 40px;
+  }
+
+  &__title {
+    font-family: $main-font-family;
+    font-size: 12px;
+    font-weight: $medium;
+    line-height: 18px;
+    color: $primary-color;
+    align-content: space-evenly;
```
TarasHoliuk: redundant

src/styles/card.scss
```diff
@@ -0,0 +1,114 @@
...
+    color: $primary-color;
+    align-items: end;
+    margin: 0;
+  }
+
+  &__button {
+    display: block;
+    text-align: center;
+    box-sizing: border-box;
+    width: $main-width;
```
TarasHoliuk, line 78: just 100% instead of hardcoded value in px

src/styles/header.scss
```diff
@@ -0,0 +1,16 @@
+body {
```
TarasHoliuk: shouldn't be here

TarasHoliuk (changes requested):

src/styles/header.scss
etojeDenys: It's the general styles, that should be in the index.scss
image

src/index.html
etojeDenys: don't forget to fix it
image

src/index.html
etojeDenys: image

etojeDenys (changes requested):

PR #5105, opened 2025-02-07
created catalog

kostivkostiv:
- [DEMO LINK](https://kostivkostiv.github.io/layout_catalog/)
- [TEST REPORT LINK](https://kostivkostiv.github.io/layout_catalog/report/html_report/)

- [X] All components follow BEM and use SCSS
- [X] Repeated sizes and special colors are put to variables
- [X] Cards are shown in 1, 2, 3, or 4 columns based on screen resolution

### DIFF
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -1,4 +1,4 @@
-@import "./variables";
+@import './variables';
 
 html {
   font-family: Roboto, Raleway, sans-serif;
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -1,6 +1,4 @@
-:root {
-  --main-color: #00acdc;
-}
+@import "./variables";
 
 html {
   font-family: Roboto, Raleway, sans-serif;
@@ -76,7 +74,7 @@ body {
 
 .is-active {
   position: relative;
-  color: var(--main-color);
+  color: $main-blue-accent-color;
 
   &::after {
     content: '';
@@ -85,15 +83,7 @@ body {
     width: 37px;
     height: 4px;
     flex-shrink: 0;
-    background-color: var(--main-color);
+    background-color: $main-blue-accent-color;
     border-radius: 8px;
   }
-}
-
-[data-qa='hover'] {
-  color: #000;
-}
-
-[data-qa='hover']:hover {
-  color: var(--main-color);
 }
--- a/src/styles/header.scss
+++ b/src/styles/header.scss
@@ -1,0 +1,99 @@
+:root {
+  --main-color: #00acdc;
+}
+
+html {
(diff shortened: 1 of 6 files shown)


src/styles/header.scss
```diff
@@ -0,0 +1,89 @@
...
+  background: #fff;
+
+  &__logo {
+    display: flex;
+    align-items: center;
+
+    &-img {
+      width: 40px;
+      height: 40px;
+      background: lightgray 50% / cover no-repeat;
```
alexander-ignatow, line 29: minor - but this is redundant:

```suggestion
      background: lightgray 50% ;
```

Cover and no-repeat won't have any effect on img src
If you want to control how the image  (set by src) is positioned - check out `object-position` and `object-fit`  CSS properties. For instance:

```
object-fit: cover;
```

src/styles/header.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.is-active {
+  position: relative;
+  color: $main-blue-accent-color;
+
+  &::after {
+    content: '';
+    position: absolute;
+    bottom: 0;
+    width: 37px;
```
alexander-ignatow, line 83: Don't set `width` explicitly.

You need to be able to re-use the same style on the other menu elements with different widths.

You can try following (as long as your `after` is positioned absolutely):
```suggestion
    left: 0;
    right: 0;
```

src/styles/header.scss
```diff
@@ -0,0 +1,89 @@
...
+
+html {
+  font-family: Roboto, Raleway, sans-serif;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
alexander-ignatow, line 11: this is inconsistent with `main.scss`

In your case it make sense to put all general styling (for HTML, body, p, a, etc) in one file

src/styles/main.scss
```diff
@@ -0,0 +1,164 @@
...
+    box-sizing: border-box;
+  }
+
+  &__card:hover {
+    transform: scale(120%);
+  }
+
+  &__card:hover .main__paragraph1 {
+    color: #34568b;
+  }
```
alexander-ignatow, line 163: Nest this please

```suggestion
  &__card:hover {
    transform: scale(120%);
 
  & .main__paragraph1 {
    color: #34568b;
  }
  }
```

alexander-ignatow (approved):
Good job, but several minor issues are there

Let's move on, but please read all my comments and make sure to avoid these errors in the future!
Make sure you use BEM correctly, as the incorrect usage kills all pros of the approach and causes a lot of problems.

If you have any questions don't hesitate and hop in the QnA!

PR #5201, opened 2025-03-18
add task solution

souzasfelipe:
[DEMO LINK](https://souzasfelipe.github.io/layout_catalog/)
[TEST REPORT LINK](https://souzasfelipe.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -17,11 +17,407 @@
     />
     <link
       rel="stylesheet"
-      href="styles/index.scss"
+      href="styles/index.css"
```
pedro-ruas, line 20: Is this correct? I don't see any styling applied in the Demo

pedro-ruas (changes requested):
You should double-check the styling files, because your demo isn't working and it's not passing any test.
By the looks of it, seems like you are trying to use the wrong styling file

PR #5282, opened 2025-04-13
add task solution

stanislav-sn:
- [DEMO LINK](https://stanislav-sn.github.io/layout_catalog/)
- [TEST REPORT LINK](https://stanislav-sn.github.io/layout_catalog/report/html_report/)


stanislav-sn:
@mateacademy-ai-mentor 
I've made the changes you mentioned, but: 
1) I have "data-qa='card'" only in the first card;
2) you say that "scale(1.2)" affects neighbouring elements, and at the same time you say that I should use "transform: translate", but "transform: translate" does not enlarge the element, but shifts it, so I correctly applied "scale(1.2)".

PR #5406, opened 2025-05-19
add task solution

sidond:
- [DEMO LINK](https://sidond.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sidond.github.io/layout_catalog/report/html_report/)


src/index.html
```diff
@@ -20,8 +21,404 @@
...
+          class="logo"
+        />
+      </a>
+
+      <nav class="nav-links">
+        <ul class="nav-items">
+          <li class="nav-item">
+            <a
+              href="#apple"
+              class="nav-link nav-link--is-active"
```
TarasHoliuk, line 45:
```suggestion
              class="nav-link nav-link is-active"
```
Add just `is-active` as described (in one of the previous tasks, at least, about Header only). It's needed to pass all tests (test tries to find element with `is-active` class)

src/styles/main.scss
```diff
@@ -0,0 +1,38 @@
...
+  font-family: Roboto, sans-serif;
+  color: $default-text-color;
+
+  font-optical-sizing: auto;
+  font-weight: 500;
+  font-style: normal;
+  font-variation-settings: 'wdth' 100;
+}
+
+.product-grid {
```
TarasHoliuk, line 13: in fact, it's a separate block

src/index.html
```diff
@@ -20,8 +21,404 @@
...
+      >
+        <img
+          src="images/logo.png"
+          alt="Moyo logo"
+          class="logo"
+        />
+      </a>
+
+      <nav class="nav-links">
+        <ul class="nav-items">
```
TarasHoliuk, line 41:
```suggestion
        <ul class="nav__list">
```
use correct BEM naming (example means that nav is a separate BEM block so its styles should be in a separate file)

src/index.html
```diff
@@ -20,8 +21,404 @@
...
+        class="logo-link"
+      >
+        <img
+          src="images/logo.png"
+          alt="Moyo logo"
+          class="logo"
+        />
+      </a>
+
+      <nav class="nav-links">
```
TarasHoliuk, line 40:
```suggestion
      <nav class="nav">
```

src/index.html
TarasHoliuk: use correct BEM naming everywhere

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,64 @@
+.header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 50px;
+  text-transform: uppercase;
+  box-shadow: 0 2px 4px 0 #0000000d;
+}
+
+.logo-link {
```
TarasHoliuk, line 10: make it an element of the header block (BEM naming) and use SCSS nesting

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,120 @@
...
+    &__product-code {
+      margin-top: 4px;
+      font-size: 10px;
+      line-height: 14px;
+      font-weight: $light-font;
+      color: $lighter-text-color;
+    }
+  }
+
+  .stars {
```
TarasHoliuk, line 38: it's a separate block - move to a separate file

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,120 @@
...
+
+  transition: transform 300ms;
+
+  &__photo {
+    margin-top: 16px;
+    width: 160px;
+    height: 134px;
+  }
+
+  .description {
```
TarasHoliuk, line 19: it also should be BEM element (of the card block)

src/styles/blocks/product-card.scss
```diff
@@ -0,0 +1,120 @@
...
+
+      margin-right: $stars-gap;
+
+      &:nth-child(-n + 4) {
+        background-image: url(../images/star-active.svg);
+      }
+    }
+  }
+
+  .rating {
```
TarasHoliuk, line 56: make it an element of card block. same for `price-string` (bad name BTW - use `card__price-wrapper`, etc)

TarasHoliuk: Note: you can have nested into each other elements

For example, card__rating can be an element of the card block, and card__reviews can be an element of the card block. And it's allowed you to put card__reviews inside of card__rating in HTML structure

TarasHoliuk (changes requested):
GJ 👍  Added a few comments

PR #5408, opened 2025-05-19
Catalog 

BorisMund:
- [DEMO LINK] - (https://borismund.github.io/layout_catalog/) 
- [TEST REPORT LINK] - 


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. The first link looks different as on the design 
image

2. Don't need to change the background color of button when hovering the full card, need to change background color only when hovering button 
image

3. Need to change color of these links when hovering 
image

PR #5453, opened 2025-06-15
Develop

Yetzir:
https://leospartadev.github.io/layout_catalog/

https://leospartadev.github.io/layout_catalog/report/html_report/


etojeDenys (approved):
Good job!
However, it would be better to change the text color on hover

image

PR #5705, opened 2025-09-11
update

ded7g:


etojeDenys (changes requested):
please provide demo and test links, and request a review again

PR #5775, opened 2025-10-17
solution

stager0:


etojeDenys (changes requested):
provide demo and test links, then request a review again

PR #5878, opened 2025-11-29
add solution

ALendyel-749:


etojeDenys (changes requested):
1. test link does not work

2. hover effect looks wrong
image

please fix these points and request a review again

etojeDenys (changes requested):
try to pass all the tests
image

PR #5938, opened 2025-12-16
Develop

sanchez-primal:
- [DEMO LINK](https://sanchez-primal.github.io/layout_catalog/)
- [TEST REPORT LINK](https://sanchez-primal.github.io/layout_catalog/report/html_report/)

Hey @mateacademy-ai-mentor, I'm not done with the project, but I can't figure out what's the problem the BackstopJS tests are pointing to. It looks like there's a pixel-perfect difference in a card block, but I've copied the code from a separate project where I only had to do the card and it passed the tests there. 
Don't mind pointing out other flaws yet, I'm planning on finish everything else myself as soon as I'm done with pixel-perfection.


sanchez-primal:
@mateacademy-ai-mentor, I'm checking in for the day. Fixed animations and gaps, did not yet put all values into variables, and still can't pass all the BackstopJS tests despite having copied the card element from a successful project. Can you help me figure out what's wrong with cards? The text is just a bit different here and there.

sanchez-primal:
@mateacademy-ai-mentor there are some variables that are only used once, why should we make them variables and not leave them hardcoded as an additional reminder that "this value is not bound to anything"? It will not be hard to go to that place and change it, as it also will not affect anything.

sanchez-primal:
@mateacademy-ai-mentor 
Regarding my question about variables: I don't exactly understand how is this single source of truth easier to use than when it doesn't have the single use values, IMO it just makes the codebase more complex. Can you explain it in greater detail and provide some simple examples, or at least one?

PR #6011, opened 2026-01-29
add task solution

inklynx:
- [DEMO LINK](https://inklynx.github.io/layout_catalog/)
- [TEST REPORT LINK](https://inklynx.github.io/layout_catalog/report/html_report/)


inklynx:
@mateacademy-ai-mentor
Regarding unitless value (line-height factor): Your suggestion is good, wise and practical. In my own project I would use it right away. However, here I'm dealing with many line-heights - each depends on the element (in one this factor is 1.4, in another 1.5, in yet another place these are some creepy values). This is how it is set in the reference, while the pixel-perfect test checks the bitmap. The author of the reference hardcoded the line spacing values because his layout would also break in these conditions (like mine does now, so it won't pass the BackstopJS test). I wrote a function that counts the dependency ratio between the font size and the line height, but this seems very impractical in a situation where the code is not isolated but works in browser conditions. What will you advise me in that situation?

PR #6381, opened 2026-08-02
add solution

Whitefyre:
- [x] All changes on :hover are smooth
https://whitefyre.github.io/layout_catalog/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,7 +9,7 @@ Create an HTML page with a catalog. Develop semantic page structure as shown on 
 - add `data-qa="card-hover"` (not just `hover`) to the link `Buy` inside the first card
 - nav links color is not `black` anymore (nav links should have `#060b35` color)
 - add the class `is-active` to the first link (`Apple`) in the navigation
-- use `<main>` tag for cards container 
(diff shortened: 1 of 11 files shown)


Whitefyre:
@mateacademy-ai-mentor i do everything as you said, except is-active class. The requirements state that it must be there, so I'm leaving it in

brespect (changes requested):
Good progress, looks good but you need 8 items against 4 and try to pass all tests before requesting mentor's review

Whitefyre:
I fixed some issues with the tests by adding the line-height from the Figma mockup to &__title, &__product-code, and &__price-text, then added 4 more cards. However, theres still a minor issue with the “Buy” button thats causing everything to shift slightly. I compared the Figma properties with mine and can’t figure out what the problem might be. Please approve the task, I don't think its that critical.

Anton-Kuchmasov (approved):
For this time, it's OK, but please don't ignore Figma mockup & tasks requirements in your further exercises - if there 8 (eight) cards in the Grid, there should be EXACTLY 8 (eight) cards

